PR #7, opened 2017-11-27
feat: switch to stderr and default port

bcoe:
allow inspector to pick a default port and, rather than polling for port, wait until port # is output to `stderr` 

I think if we were to add a standardized prefix to debugger output in Node.js, we could easily strip any messages generated by the debugger, while still kicking along other messages to `stderr` (refs: https://github.com/nodejs/node/issues/17337).

---

@TimothyGu, I assume that this message isn't output until the port is taking connections?

reviewer: @TimothyGu

### DIFF
--- a/bin/c8.js
+++ b/bin/c8.js
@@ -2,20 +2,13 @@
 
 const argv = require('yargs').parse()
 const CRI = require('chrome-remote-interface')
-const getPort = require('get-port');
-const foreground = require('foreground-child')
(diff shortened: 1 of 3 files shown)


TimothyGu:
> I assume that this message isn't output until the port is taking connections?

That's right.

----

Also, this technique is used by the official `node inspect` command as well, so it should be well supported.

PR #8, opened 2017-12-01
feat: implement Istanbul reporting

bcoe:
using various bits of nyc as a foundation this pull request implements outputting v8 coverage information through Istanbul's reporters.

As an example, the following:

```bash
./bin/c8.js --reporter html test/fixtures/a.js 
```

Outputs an HTML report in `./coverage/index.html` that looks like this:

screen shot 2017-12-01 at 12 13 23 am

I recommend running a [canary version of Node.js](https://github.com/v8/node) to see all the fancy new coverage features in action.

CC: @schuay, @JaKXz, @isaacs


bin/c8.js
```diff
@@ -42,24 +62,40 @@ const spawn = require('../lib/spawn')
...
-    return isAbsolute(url) && IGNORED_PATHS.every(ignored => !ignored.test(url))
+    url = url.replace('file://', '/') // ES6 modules.
+    return isAbsolute(url) && exclude.shouldInstrument(url)
+  })
+  return result
+}
+
+function writeIstanbulFormatCoverage (allV8Coverage) {
+  const tmpDirctory = resolve(argv.coverageDirectory, './tmp')
+  rimraf.sync(tmpDirctory)
```
TimothyGu: Not like it matters much, but can you use asynchronous I/O if possible?

PR #11, opened 2017-12-18
chore: add support for subprocess spawning

bcoe:
using the spawn-wrap library, this pull request adds support for tracking coverage in applications that spawn subprocesses (this is important in unit testing since many test frameworks immediately spawn a subprocess for executing tests).

_I was shocked when I pulled this feature together by the fact that everything *just worked*_

This raises a couple mysteries:

## Mystery 1

I'm able to output coverage information in the `process.on('exit')` handler:

```js
    // hook process.exit() and common exit signals, e.g., SIGTERM,
    // and output coverage report when these occur.
    onExit(() => {
      session.post('Profiler.takePreciseCoverage', (err, res) => {
        if (err) console.warn(err.message)
        else {
          try {
            const result = filterResult(res.result)
            writeIstanbulFormatCoverage(result)
          } catch (err) {
            console.warn(err.message)
          }
        }
      })
    }, {alwaysLast: true})
```

I was pleasantly surprised to find out that I'm able to call `Profiler.takePreciseCoverage` from within an exit handler; this is a bit shocking to me because you're generally not able to perform async operations within this context (was a workaround for this introduced, in relation to the inspector?).

## Mystery 2

block-level test coverage seems to be working on the canary build of Node.js:

screen shot 2017-12-17 at 4 24 45 pm

☝️ I'm ecstatic that this is working; but shouldn't it not be working since I'm not spawning with `--inspect-brk`? I was under the impression once the Node process starts we're not able to enable block level coverage.

---

@hashseed, @TimothyGu @schuay @eugeneo, perhaps you can provide some insight into these mysteries?

If things are working as expected, it would seem that we're at the point where we can use v8's built-in test coverage in Node.js pretty effectively 🎉


schuay:
> :point_up: I'm ecstatic that this is working; but shouldn't it not be working since I'm not spawning with --inspect-brk? I was under the impression once the Node process starts we're not able to enable block level coverage.

I'm not sure what the specific timing is in your case. As long as block coverage is enabled prior to a function being parsed, we can provide block-granularity data. This *could* also be due to things like lazy parsing of non top-level functions. Awesome that it works here :)

lib/wrap.js
```diff
@@ -0,0 +1,81 @@
...
+const exclude = Exclude({
+  include: argv.include,
+  exclude: argv.exclude
+})
+
+;(async function runInstrumented () {
+  try {
+    // bootstrap the inspector before kicking
+    // off the user's code.
+    inspector.open(0, true)
```
TimothyGu: This isn’t necessary. The in-process Inspector implementation works without launching the WebSocket server, which is what `inspector.open` does.

lib/wrap.js
```diff
@@ -0,0 +1,81 @@
...
+  }
+})()
+
+function filterResult (result) {
+  result = result.filter(({url}) => {
+    url = url.replace('file://', '')
+    return isAbsolute(url) &&
+      exclude.shouldInstrument(url) &&
+      url !== __filename
+  })
```
TimothyGu: This filtering mechanism generally doesn’t work very well for Windows, but I don’t really have any easy way to fix that. I’ve been wanting to publish a module that does this conversion the same way Node.js core does it, and I’ll report back when I do.

PR #13, opened 2018-07-29
chore: update dependencies

demurgos:
This commit updates the dependencies to their latest version.

The only difference is a different expected output for the integration tests.


test/integration.js
```diff
@@ -15,12 +15,12 @@ describe('c8', () => {
       cwd: process.cwd()
     })
     output.toString('utf8').should.include(`
-------------|----------|----------|----------|----------|----------------|
-File        |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
-------------|----------|----------|----------|----------|----------------|
-All files   |    91.67 |      100 |       80 |    91.67 |                |
- normal.js  |    85.71 |      100 |       50 |    85.71 |       14,15,16 |
- timeout.js |      100 |      100 |      100 |      100 |                |
-------------|----------|----------|----------|----------|----------------|`)
```
TimothyGu: Are these changes expected?

demurgos: I had these results even without updating the dependencies, so I assumed that the change was caused by simply having a more recent version of Node.

By looking into it deeper, you are right that it reveals an issue. The functions in `fixtures/normal` are counted as `0/1` instead of `1/2`. The `apple` function is not detected as a function.

I should probably revert the change for this file: the tests were already failing before updating the dependencies and there's no CI anyway.

demurgos:
I isolated the example in a minimal setting. I currently get odd coverage results (not all the functions are matched) that may be related to the changes here. I need to check if there's an issue with the way I test it or if it's a Node/V8 issue.

demurgos:
Ok, I managed to get the right order of invocations to get V8 to return the coverage for `normal.js` with all the functions. I'm now check with `v8-to-istanbul` if it is properly converted.

demurgos:
My refactored `v8-to-istanbul` can now properly handle the functions (and statements) in `normal.js`. Remaining work include supporting branches and integrating my changes into `c8`.

Issue #14, opened 2018-07-29
Multiple incompatible reports generated for the same file.


demurgos:
I am investigating a bug where the `lcovonly` reporter sometimes crashes. I cannot reproduce it reliably, but it happens about 60% of the time.

I was able to find part of the cause why it happens, but I need help to find the root cause and solve it.

The `lcovonly` report failed [at this line](https://github.com/istanbuljs/istanbuljs/blob/829e658dfa91e3a9533842be9ce940dbe7785c09/packages/istanbul-reports/lib/lcovonly/index.js#L52) because the `branchMap` and `b` properties of the file coverage had different keys. It is a broken invariant, meaning that the coverage map was malformed.

The bad file coverage object was for a file called `test.esm.js`. When looking into c8's tmp directory (`coverage/tmp`), I found two coverage file for `test.esm.js`: [gist](https://gist.github.com/demurgos/b1b1a92f89b9581b004fc700a40c3462).
If you look at them, you see that they have different `branchMap` values. The first one has 2 properties, the second one has 4 properties.

When `c8` builds its coverage map, it merges all the files in its temporary directory. One of the preconditions of the merge function is that if there are two reports for the same file, the file needs to have the same structure (ex. `branchMap`). [https://github.com/istanbuljs/istanbuljs/blob/829e658dfa91e3a9533842be9ce940dbe7785c09/packages/istanbul-lib-coverage/lib/file.js#L246](https://github.com/istanbuljs/istanbuljs/blob/829e658dfa91e3a9533842be9ce940dbe7785c09/packages/istanbul-lib-coverage/lib/file.js#L246).

---

So here is the current state of my investigation: `lcovonly` breaks because it gets an invalid file coverage because multiple reports for the same file but with different structures are merged, thus breaking one of the preconditions of the merge function.

To solve this, c8 should ensure that the reports are unique for each file URL, and that they have the same structure.

demurgos:
In [the code emitting the test reports](https://github.com/bcoe/c8/blob/f14508e0a1a2625252193ecc4c6500961cc6a7a2/lib/wrap.js#L73), I added `console.log(v8.url)` and got:

```
/data/projects/various/turbo-gulp/build/test/test.esm.js
file:///data/projects/various/turbo-gulp/build/test/test.esm.js
file:///data/projects/various/turbo-gulp/build/test/e2e/node-lib/node-lib.spec.mjs
file:///data/projects/various/turbo-gulp/build/test/e2e/node-lib/meta.js
[...]

PR #15, opened 2018-07-29
fix: filter out results without `file` protocol

demurgos:
This commit ensures that the URL of the results starts with `file://`. Some files produce an extra result without the `file` protocol, this result interferes then with report merging.

See bcoe/c8#14


demurgos:
This PR breaks c8 for CJS files, it needs more research to understand the issue/fix.

PR #17, opened 2018-08-02
fix: make tests run on Windows

laggingreflex:


test/integration.js
```diff
@@ -7,7 +7,8 @@ require('chai').should()
 
 describe('c8', () => {
   it('reports coverage for script that exits normally', () => {
-    const {output} = spawnSync(c8Path, [
+    const {output} = spawnSync('node', [
```
demurgos: It should use `process.execPath` instead of `node`.
As is, this would fail if `node` is not in the `PATH`. (It may also cause subtle bugs if there are multiple versions of Node)

laggingreflex: Thanks! fixed.

bcoe:
@laggingreflex see #22 let's fix some of the outstanding issues once we switch to using Node's built in coverage 👍

PR #18, opened 2018-08-02
fix: tests

laggingreflex:


test/integration.js
```diff
@@ -18,9 +18,9 @@ describe('c8', () => {
 ------------|----------|----------|----------|----------|----------------|
 File        |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
 ------------|----------|----------|----------|----------|----------------|
-All files   |    91.67 |      100 |       80 |    91.67 |                |
- normal.js  |    85.71 |      100 |       50 |    85.71 |       14,15,16 |
- timeout.js |      100 |      100 |      100 |      100 |                |
+All files   |    91.67 |       80 |        0 |    91.67 |                |
+ normal.js  |    85.71 |       75 |        0 |    85.71 |       14,15,16 |
```
laggingreflex: Not sure why it was 100% branch coverage, [this line in `normal.js`](https://github.com/bcoe/c8/blob/master/test/fixtures/normal.js#L9) should never have lead to 100%:

```js
 if (true || false) {
```

demurgos:
Related discussion: https://github.com/bcoe/c8/pull/13#pullrequestreview-141365295

There is still an issue here: the new version has `0/1` (0%) covered functions instead of `1/2` (50%).

bcoe:
see #22

PR #22, opened 2018-09-08
feat: switch to using Node's built in coverage

bcoe:
switches to using [Node's built in coverage](https://github.com/nodejs/node/pull/22527), which we can enable by setting `NODE_V8_COVERAGE` rather than kicking off an inspector session.

screen shot 2018-09-08 at 3 51 40 pm

_I'm beyond thrilled by how much better this implementation is working._

BREAKING CHANGE: switches to using NODE_V8_COVERAGE

CC: @demurgos, @schuay


bcoe:
CC: @addaleax, @bengl, @boneskull, @hashseed, @evanlucas

lib/parse-args.js
```diff
@@ -28,6 +28,10 @@ yargs
     default: './coverage',
     describe: 'directory to output coverage JSON and reports'
   })
+  .option('resolve', {
+    default: '',
+    describe: 'resolve paths to alternate base directory'
```
bcoe: this will be required for the Node.js codebase itself.

bcoe:
see also https://github.com/bcoe/v8-coverage-merge, which handles merging coverage from multiple inspector sessions (this will be necessary for Node.js' test suite).

lib/report.js
```diff
@@ -25,9 +40,34 @@ class Report {
...
+      const result = mergedResults[url]
+      // console.info(JSON.stringify(result, null, 2))
+      try {
+        const path = resolve(this.resolve, result.url)
+        const script = v8toIstanbul(path)
+        script.applyCoverage(result.functions)
+        map.merge(script.toIstanbul())
+      } catch (err) {
+        // most likely this was an internal Node.js library.
+        if (err.code !== 'ENOENT' && err.code !== 'EISDIR') throw err
```
bcoe: @TimothyGu any thoughts about how to easily tell whether the coverage outputted was for one of Node.js' built in libraries, e.g., `internal/inspector`.

demurgos:
This sounds great but I'll have to take a deep look at it.
It's nice to have a solution natively supported by Node, but using the inspector session also has advantages.
`spawn-wrap` is awful but it enables the cleanest `c8` implementation IMO: no temporary files, no concurrence issues (can run multiple coverage at once), enables a Node API (lib usage), no messing with exit handlers, etc.
Using the env variable (`NODE_V8_COVERAGE`) is more reliable than spawn wrap, but it brings back a few issues (mainly the fact that the `c8` cli and covered process cannot communicate directly but through files stored in a temporary directory).

(This is just a quick comment after reading the first message and checking the Node PR to add `NODE_V8_COVERAGE`, I did not look at the details yet.)

bcoe:
@demurgos spawn-wrap and signal-exit are end of life and will not be supported any longer now that this functionality has landed in Node.js itself.

Behind the scenes Node.js is kicking off the same inspector session that c8 was spawning, but in a much more reliable manner because it happens before the isolate executes and there is no need to pause the inspector; the inspector dance was bug prone, didn’t work for a variety of edge-cases (non-zero application exit codes being a major issue, which c8 handles even worse than nyc).

I recommend forking c8, if you’d like to continue interacting with the inspector directly in some cases — maybe with a more explicit name like “inspector-coverage”; I’d happily link to this in the README.

bcoe:
@demurgos I'm going to take my thoughts to email with you, let's work together to make sure your hard work on the `next` branch of `c8` isn't taken for granted. Even though on `c8` itself, I'd like to migrate away from using this approach for test coverage.

lib/report.js
addaleax: How about checking whether the URL was an absolute path or not?

bcoe:
I need this for some Node.js mentorship work I'm doing tonight, so I've gone ahead and landed. I'm still open to more conversation about whether we maintain the old approach in another library.

PR #26, opened 2018-09-13
feat: add clean command

bcoe:
Adds clean command, which allows the deletion of the temporary directory to be skipped between runs of c8.

This is useful for our own test coverage, each run of c8 was stepping on the other's toes.

PR #28, opened 2018-09-13
feat: improve test assertions

laggingreflex:
(this is on top of #26)

When a test fails, it shows an output like this, which is hard to look at:

```
  1) c8
       merges reports from subprocesses together:
     AssertionError: expected ',first\n\nsecond\n\n--------------------|----------|----------|----------|----------|-------------------|\nFile
             |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |\n--------------------|----------|----------|----------|----------|----------
     ---------|\nAll files           |    94.12 |    73.08 |        0 |    94.12 |                   |\n bin                |    83.72 |    57.14 |      10
```

This commit tries to improve the error message by using snapshot testing.


laggingreflex:
@JaKXz Good idea, thanks! Updated. 

One test that uses `should.match(regex)` assertion couldn't be used with snapshot though.

bcoe:
@laggingreflex @JaKXz I like this too; shall we land this instead of my `clean` branch? looks like you already rebased.

bcoe:
@laggingreflex oh, one suggestion. Can we add an npm command that takes the snapshots?

laggingreflex:
Sure, added `test:snap`

Issue #9, opened 2017-12-05
fun: switch to async IO


bcoe:
c8 already requires a bleeding edge v8 engine, we should switch to `async`/`await` rather than using any synchronous IO.

see @TimothyGu's comments in https://github.com/bcoe/c8/pull/8.

bcoe:
@Bnaya 👋I'm not a huge advocate of Typescript; part of why I'm excited about native coverage support in v8 is that it can eliminate some of the transpilation that is currently being performed by Istanbul.

But! I would love your help on c8; if you'd like to contribute, strike up a conversation with me in this Slack:

http://devtoolscommunity.herokuapp.com/

ProfNandaa:
@bcoe - would like to take up this one (for fun :)). I can't seem to get the code that is being referred here:

> see @TimothyGu's comments in #8.

in `bin/c8.js`, do we rewrite `rimraf` (has Async API) and `mkdirp` (no async, so means rewriting with `fs.mkdir`)?

Do we also change the `fs.*Sync`'s in `/lib/parse-args.js` and `/lib/report.js` too?

bcoe:
@ProfNandaa I'd have a preference with just using the new `fs.mkdir`, but this is something we could patch in Node.js itself; just switch the command used in Node's coverage to set recursive `true`.

I think we could switch everything that's blocking to `async`/`await`, with some fiddling with flow control -- the main advantage I see to doing this is that Mozilla's source-map library is async now:

https://github.com/mozilla/source-map

ProfNandaa:
@bcoe -- I'll be on it, hopefully should raise the PR tomorrow (going to bed now).

Thanks for the heads-up about the Mozilla project, checking it out 👍

bcoe:
@profnandaa thank you for this contribution.

PR #32, opened 2018-09-18
chore(bin): rewrite to async I/O

ProfNandaa:
- Replace mkdirp with the new fs.mkdir with `recursive` option
- Rewrite all top level IO to async (mkdir and rimraf)

Remaining work: to rewrite `lib/*.js` to async I/O (needs discussion)

Ref: #9


ProfNandaa:
Don't land yet, until the `mkdirp` vs. feature-detection issue is resolved and the final API shipped. re: https://github.com/nodejs/TSC/issues/601

Thanks @addaleax for the review! 🙏

bin/c8.js
```diff
@@ -40,3 +29,20 @@ function outputReport () {
...
 }
+
+(async function run () {
+  if (argv._[0] === 'report') {
+    argv = yargs.parse(process.argv) // support flag arguments after "report".
+    outputReport()
+  } else {
+    if (argv.clean) {
+      await _p(rimraf)(argv.tempDirectory)
+      await _p(fs.mkdir)(argv.tempDirectory, { recursive: true })
```
bcoe: unfortunately this functionality will not yet work: https://github.com/nodejs/node/pull/21875

bin/c8.js
```diff
@@ -40,3 +29,20 @@ function outputReport () {
...
   })
 }
+
+(async function run () {
+  if (argv._[0] === 'report') {
+    argv = yargs.parse(process.argv) // support flag arguments after "report".
+    outputReport()
+  } else {
+    if (argv.clean) {
+      await _p(rimraf)(argv.tempDirectory)
```
bcoe: @addaleax @jasnell do we know the timing of when the `fs.promises` API will stop being experimental? Given that this library is exercising various bleeding-edge features, e.g., the Inspector itself, my preference would be for using `fs.promises` rather than `util.promisify`.

ProfNandaa: @bcoe - still `fs.promises` experimental though, can I just stick to at least the stable(r) ones for now? 🤔 So that the only thing we have to worry about right now is the recursive `fs.mkdir` shipping in the next + 1 release.

bcoe: sounds good to me, let's stick with util.promisify ... I'm excited for fs.promises though!

ProfNandaa:
Good news - https://github.com/nodejs/node/pull/21875#issuecomment-424156156 / Once this is done, we'll land this too.

demurgos:
Recursive `mkdir` will be in Node 10.12, staged to be released soon.

bin/c8.js
jasnell: fs.promises will likely go stable early in 11.x timeframe (hopefully before end of the year)... We should have at least one module in CITGM that uses it before it can graduate.

ProfNandaa: @jasnell -- good to hear! Happy to move over to `promises` then since `fs.mkdir` is now out of question. I'll update the PR.

demurgos:
@ProfNandaa Don't forget to update `engines.node` to require the version with recursive `mkdir` before merging.

ProfNandaa:
@demurgos &mdash; good call, thx.

bcoe:
@ProfNandaa awesome work!

PR #33, opened 2018-09-20
docs: add clarification about goals of this project

bcoe:
@demurgos and I have been having long conversations about implementation details surrounding c8, and I think part of the problem was that I've done a poor job of describing the goals I have in mind (they lived only in my head, which isn't useful).

tldr; I'd like this project to demonstrate a powerful code coverage solution relying primarily on native V8 functionality, and minimal user-land modules.

_data-transformations are okay, but I don't want to parse JavaScript with Babel, muck with Node's runtime, etc._

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,4 +32,4 @@ A fully functional code coverage solution using only V8's native coverage
 features and minimal user-land modules:
 
 * no parsing of JavaScript code.
-* no mucking with Node's runtime environment.
+* no mucking with Node.js' runtime environment.
--- a/README.md
+++ b/README.md
@@ -20,8 +20,16 @@ The above example will output coverage metrics for `foo.js`.
 
 run `c8 report` to regenerate reports after `c8` has already been run.
 
-## Disclaimer
+## Supported Node.js Versions
 
 c8 uses
 [bleeding edge Node.js features](https://github.com/nodejs/node/pull/22527),
 make sure you're running Node.js `>= 10.10.0`.
+
+## Goals of Project
+
+A fully functional code coverage solution using only V8's native coverage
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,8 +20,16 @@ The above example will output coverage metrics for `foo.js`.
...
+## Supported Node.js Versions
 
 c8 uses
 [bleeding edge Node.js features](https://github.com/nodejs/node/pull/22527),
 make sure you're running Node.js `>= 10.10.0`.
+
+## Goals of Project
+
+A fully functional code coverage solution using only V8's native coverage
+features and minimal user-land modules:
```
JaKXz: Being strict here, this sentence does not really tell me what the following list is for. I would add:

`...modules, so that we fit these constraints:` or something to that effect.

README.md
```diff
@@ -20,8 +20,16 @@ The above example will output coverage metrics for `foo.js`.
...
 run `c8 report` to regenerate reports after `c8` has already been run.
 
-## Disclaimer
+## Supported Node.js Versions
 
 c8 uses
 [bleeding edge Node.js features](https://github.com/nodejs/node/pull/22527),
 make sure you're running Node.js `>= 10.10.0`.
+
+## Goals of Project
```
JaKXz: nit: Goals of `this` Project or Goals of `the` Project.

README.md
ProfNandaa: Glad to make this change 👍

README.md
ProfNandaa: On it 👍

demurgos:
Looks good to me, thanks for adding this!

PR #45, opened 2018-11-20
fix: Windows file:\/\// -> file:[\/]+

### LINKED ISSUE #41: `c8 report` doesn't work on Windows
* **Version**:
3.2.1
* **Platform**:
Windows 10

`c8 report` always outputs an empty report block on windows, no matter what `include` or `exclude` options you do (or don't) provide. Looking at it, it looks like the coverage files being produced on windows have paths like `file:///E:/Github/TypeScript/built/local/tsc.js`, which `c8` then shortens to `/E:/Github/TypeScript/built/local/tsc.js` internally, which I'm guessing don't glob match (anything) correctly inside `test-exclude`, since that's not how windows-style paths are typically rooted (though I'm not sure - that last bit's a guess, since I'm not too familiar with the package).

laggingreflex:
On Windows (edit: [it seems](https://superuser.com/questions/352133/why-do-file-urls-start-with-3-slashes) this can be a case on other OSes too) there seems to be `file:///` i.e. 3 forward slashes, whereas here it was only accounting for 2, leaving in an extra `/`

Fixes #41

##### Checklist

   Weren't passing for me before this fix either.
 
~~`npm run test:snap` (to update the snapshot)~~ shouldn't be needed
~~tests and/or benchmarks are included~~
~~ documentation is changed or added~~

### DIFF
--- a/lib/report.js
+++ b/lib/report.js
@@ -113,7 +113,7 @@ class Report {
     const result = []
     for (const v8ScriptCov of v8ProcessCov.result) {
       // TODO: implement real fix from https://github.com/nodejs/node/issues/23783
-      v8ScriptCov.url = v8ScriptCov.url.replace(/^file:\/\//, '')
+      v8ScriptCov.url = v8ScriptCov.url.replace(/^file:[\/]+/, '')
       if (this.exclude.shouldInstrument(v8ScriptCov.url) &&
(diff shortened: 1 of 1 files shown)


laggingreflex:
#46 fixes it better. Closing

PR #46, opened 2018-11-21
fix: file URL to system path conversion

demurgos:
This commit fixes the conversion from `file://` urls to system-dependent paths. It ensures that it works on Windows and with special characters.

- Closes bcoe/c8#41
- Closes bcoe/c8#45

##### Checklist

- [ ] tests and/or benchmarks are included: N/A, tested in [furi](https://github.com/demurgos/furi/tree/master/src/test)
- [ ] documentation is changed or added: N/A


demurgos:
@bcoe 
The reason why I use `is-windows` is for edge cases such as msys or cygwin.

Regarding the tests, I am not sure if it needs specific tests: enabling CI on Windows would be more relevant (this is a platform-dependent bug).

The conversion functions have their own tests in the lib: [posix](https://github.com/demurgos/furi/blob/master/src/test/posix.spec.ts), [windows](https://github.com/demurgos/furi/blob/master/src/test/windows.spec.ts). I am confident that this library can handle any edge case: it was tested on all systems (including windows UNC paths) and my I [pushed my test suite to Node core](https://github.com/nodejs/node/pull/23720/files#diff-eee81c988084556476b847eb61680ca0) so the result should match.

PR #51, opened 2018-12-21
feat: allow script wrapper length to be specified

bcoe:
##### Checklist


bcoe:
CC: @demurgos

demurgos:
Looks good to me as a quickfix to handle CJS wrappers. I would have preferred `prefixLength` or `wrapperPrefixLength` but this name is fine too and matches the `v8-to-istanbul option`.

Longer term, I am not sure if this enough: it does not seem to support mixing CJS (wrapped) and ESM (unwrapped) modules inside the same process? I think what we'd need a mapping from globs to file-specific options. (`[{file: "**/*.mjs", wrapperLength: 0}, {file: "**/*.js", wrapperLength: 74}]`). Or we may not need it at all if https://github.com/nodejs/node/issues/17396 is fixed. I prefer to not block on this and merge this PR already.

bcoe:
@demurgos I agree, this is ugly ... but probably a reasonable hack until Node.js can get rid of the wrapper?

PR #52, opened 2018-12-23
chore: enable windows tests

bcoe:
##### Checklist


bcoe:
@demurgos it appears as though Node@11 again removes the `file://` prefix, which is what is causing the tests to fail. I don't think we need the furi dependency it turns out, as this line can't get hit on Node@11.

demurgos:
@bcoe 
I thought that Node 11 was still using `file:///` (both for CJS and ESM). I'll check it by the end of the week.

bcoe:
@demurgos it seems to be you're correct, there's a more interesting issue on Windows now. If you take a look at:

https://travis-ci.org/bcoe/c8/jobs/471722552

There's a one-line discrepancy in what we report as uncovered between Linux/OSX and Windows.

demurgos:
You're right, it looks like a more complex situation. It may be worth checking if we get the same V8 reports on all systems.

.travis.yml
```diff
@@ -1,4 +1,8 @@
 language: node_js
+os:
+  - linux
+  - osx
+  - windows
 node_js:
-  - "10.10"
-after_success: npm run coverage
\ No newline at end of file
+  - "11"
```
JaKXz: this could just be `"stable"`

JaKXz: TIL this is wrong! See https://github.com/istanbuljs/nyc/pull/961

PR #57, opened 2019-01-20
chore: enable windows tests

bcoe:
Windows tests should now work due to landing:

https://github.com/istanbuljs/v8-to-istanbul/pull/11

##### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "url": "git@github.com:bcoe/c8.git"
   },
   "scripts": {
-    "test": "node ./bin/c8.js --reporter=html --reporter=text mocha ./test/*.js",
+    "test": "node ./bin/c8.js --reporter=html --reporter=text mocha --timeout=4000 ./test/*.js",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -9,7 +9,7 @@
     "url": "git@github.com:bcoe/c8.git"
   },
   "scripts": {
-    "test": "node ./bin/c8.js --reporter=html --reporter=text mocha ./test/*.js",
+    "test": "node ./bin/c8.js --reporter=html --reporter=text mocha --timeout=4000 ./test/*.js",
```
JaKXz: Is the extra timeout just for windows?

Not to be a stickler, but, one of the really nice differentiators of this library is that it runs through tests much faster than NYC, which in turn makes development and contributions a lot easier. I would love to be in a world where that stays true as much as possible

bcoe: I had the Node.js test suite and c8 test suite running at the same time, which was slowing down my computer. Reverted this change.

bcoe:
@JaKXz scratch that, I guess Windows does take a bit longer than 2 seconds, bumped back up to 4.

PR #69, opened 2019-03-08
Add sourcemap support

dicarlo2:
Opening this PR as a rough start to adding source-map support. This works as is, but would like to get some feedback before cleaning it up, adding tests, etc.

Also, the globRoot changes are something else I needed, just didn't split them out. Would be great to get feedback on if that can be merged. If so, will split it out into another PR.

##### Checklist

### DIFF
--- a/lib/commands/check-coverage.js
+++ b/lib/commands/check-coverage.js
@@ -14,14 +14,15 @@ exports.handler = function (argv) {
   const report = Report({
     include: argv.include,
     exclude: argv.exclude,
+    globRoot: argv["glob-root"],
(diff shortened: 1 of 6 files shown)


bcoe:
@dicarlo2 thanks for this! I've admittedly been dragging my feet a bit because I'm interested in using this a testing ground for pushing for source-map support in V8 and Node.js -- I'm curious how solid this implementation is: https://github.com/v8/v8/blob/master/tools/SourceMap.js

dicarlo2:
@bcoe no problem! Admittedly we actually switched back to using nyc instrumentation because it appears to be more comprehensive than the built-in V8 instrumentation, at least currently. I'll go ahead and close for now since I don't currently need this.

PR #74, opened 2019-04-07
 test: add test for experimental ESM modules

bcoe:
Adds test demonstrating using c8 for collecting coverage for experimental modules (see conversation [here](https://github.com/nodejs/modules/issues/95)).

_I've been meaning to add this test for a while, but haven't had many OSS cycles... it's cool to see it work after months of noodling on V8/Node.js._

## Background

The new module system in Node.js does not support `require.extensions`; while there is some work in motion to support similar extension mechanisms, it's unlikely to exist in time for `--experimental-modules` to be un-flagged.

 `c8` collects coverage using the inspector and the NODE_V8_COVERAGE environment variable, this is built into Node.js itself and does not require extending loaders.

##### Checklist

CC: @MylesBorins, @joyeecheung, @guybedford, @GeoffreyBooth, @coreyfarrell (CC: party 🎉 woo).


coreyfarrell:
I tested this with https://github.com/coreyfarrell/c8-test, worked under node.js 11.13.0 but not 10.15.3.  Is c8 with experimental-modules under node.js 10 not possible?

The test repo does equivalent testing with nyc using the `esm` module for easy comparison, the results under node.js 11 seem reasonable for both.

bcoe:
@coreyfarrell it should work under 10, but you'd need to play around with `wrapper-length` I believe, might need to fiddle with a couple other things.

coreyfarrell:
@bcoe we recently saw a change in node.js 11.11.0 which switched require from `vm.runInThisContext` to `vm.compileFunction`, is this the reason the needing to fiddle with `wrapper-length`?

coreyfarrell:
Odd, if I set `"wrapper-length": 0` in the config for c8 it works for node.js 10 and 11.

bcoe:
@coreyfarrell the Node@10 issues should be addressed by the following:

https://github.com/nodejs/node/commit/23ea7ee64b98218e06cf3217372ade7b401250f8

Issue #70, opened 2019-03-14
Don't overrider NODE_V8_COVERAGE


kilianc:
Is it possible to prepopulate `--temp-directory` with the content of `$NODE_V8_COVERAGE` if set? I am trying to integrate c8 in my workflow and I really like the ability to separate execution and reporting.

* `NODE_V8_COVERAGE=/tmp/cov npm test`
* `NODE_V8_COVERAGE=/tmp/cov c8 report`

Happy to contribute id accepted.
Cheers!

shinnn:
Good idea. PR welcome.

The patch would mainly update the following two lines:

* https://github.com/bcoe/c8/blob/5ea4313c92289beed01a7fdff52243adeb50b258/lib/parse-args.js#L64 Use `process.env.NODE_V8_COVERAGE` as a fallback if it exists, instead of `'./coverage/tmp'`
* https://github.com/bcoe/c8/blob/5ea4313c92289beed01a7fdff52243adeb50b258/bin/c8.js#L29 Don't override `NODE_V8_COVERAGE` if it exists

bcoe:
@kilianc can you confirm that this is now working for you?

PR #78, opened 2019-04-20
fix: do not override NODE_V8_COVERAGE if set (#70)

kilianc:
WIP, please do not merge yet


kilianc:
How do you want me to test this properly? I think is messing up the snapshots too

test/integration.js
```diff
@@ -48,6 +48,19 @@ describe('c8', () => {
     )
   })
 
+  it('supports exeternally set NODE_V8_COVERAGE', () => {
+    process.env.NODE_V8_COVERAGE = './coverage/tmp_'
```
shinnn: Instead of overriding `NODE_V8_COVERAGE` of the whole process, use [`spawnSync()`](https://nodejs.org/api/child_process.html#child_process_child_process_spawnsync_command_args_options) option to modify only this child process' environment.

test/parse-args.js
```diff
@@ -21,6 +21,17 @@ describe('parse-args', () => {
...
+    })
+  })
+
+  describe('with NODE_V8_COVERAGE already set', () => {
+    it('should not override it', () => {
+      process.env.NODE_V8_COVERAGE = './coverage/tmp_'
+      process.argv = ['node', 'c8', '--foo=99', 'my-app', '--help']
+      const argv = buildYargs().parse(hideInstrumenteeArgs())
+      argv.tempDirectory.endsWith('/coverage/tmp_').should.be.equal(true)
+      process.env.NODE_V8_COVERAGE = ''
```
shinnn: I assume the original value is not `''`, as this test is run with `c8`.

kilianc: I can `delete` but that would create a side effect like setting it to `''`. I can use `beforeEach` and store the original value. I am only doing it that particular test now.

shinnn:
While reviewing this, I found the existing test results are too fragile to add new specs... the snapshot of the first test affects the second one, the second affects the third, and so on.

I think we should make each tests more independent before implementing new features.

@kilianc, thanks for starting this work.

kilianc:
Ok, I'll keep this parked here. @ me when I can continue!

shinnn:
> we should make each tests more independent before implementing new features.

https://github.com/bcoe/c8/pull/79

shinnn:
@kilianc Can you rebase this branch from `master`?

bcoe:
:wave: @kilianc it seems like this has stalled a bit, still a feature you need?

Could you explain your use-case for me, I don't quite understand why you'd want to run `c8` without setting the variable that allows it to start collecting coverage.

bcoe:
@kilianc this is looking pretty good to me, what work would you still like to do?

bin/c8.js
```diff
@@ -26,7 +26,10 @@ let argv = buildYargs().parse(instrumenterArgs)
       await promisify(fs.mkdir)(argv.tempDirectory, { recursive: true })
     }
 
-    process.env.NODE_V8_COVERAGE = argv.tempDirectory
+    if (!process.env.NODE_V8_COVERAGE) {
```
shinnn: Which should c8 take precedence, when both `NODE_V8_COVERAGE` environment variable and `--tmp-directory`? The current implementation will choose `NODE_V8_COVERAGE` in this case.

kilianc:
> Could you explain your use-case for me, I don't quite understand why you'd want to run c8 without setting the variable that allows it to start collecting coverage.

Quite the opposite! I am trying to give full ownership of the env to the caller, in my case CI. Instead of having to re-pass in the env as arg just to have it set again.

I hope that makes sense

kilianc:
BTW, not sure what is going on with `master` but probably someone force-pushed because I can't pull anymore. I am trying to fix my branches and force push on this PR.

kilianc:
I resolved my git issues but the test suite is still not passing. Will work on it this weekend, but I think the test suite still has some isolation problem. Cheers!

bin/c8.js
kilianc: `--tmp-directory` should win, but we default it to `NODE_V8_COVERAGE` if not set, we can get rid of this `if` statement probably

bcoe:
@kilianc once I read up on your reasoning for this feature, I totally understand what you're going for :+1: excited to see this over the finish line; let me know if there's anything I can do to help.

test/integration.js
```diff
@@ -26,6 +26,21 @@ describe('c8', () => {
     output.toString('utf8').should.matchSnapshot()
   })
 
+  it('supports exeternally set NODE_V8_COVERAGE', () => {
+    const { output } = spawnSync(nodePath, [
+      c8Path,
+      '--exclude="test/*.js"',
+      '--clean=false',
```
shinnn: > I think the test suite still has some isolation problem.

Use `--temp-directory` option here to write coverage reports to a separate directory.

bin/c8.js
shinnn: > we can get rid of this `if` statement probably

👍

test/integration.js
kilianc: @shinnn the goal of this test is indeed to use the `env` and not `--temp-directory` to set the path. Look at line 38!

kilianc:
@bcoe thanks, I think this is ready. Happy to fix anything that doesn't meet expectations :)

test/integration.js
shinnn: Ah, indeed.

bcoe:
@shinnn how are you feeling about this PR?

bcoe:
@kilianc this look reasonable to me, but all the tests appear to be failing, any thoughts?

kilianc:
@bcoe locally it was all passing, I'll try to rebase again! 🍺

bcoe:
@kilianc just wanted to bump this up, haven't unfortunately had the cycles to dig into the issues myself.

Still something you'd like to land?

kilianc:
@bcoe it is, it's just bandwidth. I'll do my best to get this done in the next week or so! Thanks for the bump :)

bcoe:
closing in favor of #122; I think there was just a bad entry in the snapshot, from prior runs of your tests.

PR #82, opened 2019-04-28
feat: force exit the process when completed

popomore:
##### Checklist

### DIFF
--- a/bin/c8.js
+++ b/bin/c8.js
@@ -27,9 +27,9 @@ let argv = buildYargs().parse(instrumenterArgs)
     }
 
     process.env.NODE_V8_COVERAGE = argv.tempDirectory
-    foreground(hideInstrumenterArgs(argv), () => {
+    foreground(hideInstrumenterArgs(argv), done => {
       outputReport(argv)
-      process.exit()
+      done()
     })
   }
 })()
--- a/bin/c8.js
+++ b/bin/c8.js
@@ -29,6 +29,7 @@ let argv = buildYargs().parse(instrumenterArgs)
     process.env.NODE_V8_COVERAGE = argv.tempDirectory
     foreground(hideInstrumenterArgs(argv), () => {
       outputReport(argv)
+      process.exit();
     })
   }
 })()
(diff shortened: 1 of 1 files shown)


addaleax:
Calling `process.exit()` without needing to is generally discouraged; for example, it may not allow scheduled writes to the disk or to stdout/stderr to finish. So, like @shinnn said, it would be helpful to know what issue this solves.

popomore:
It's unnecessary to call process.exit directly, foreground-child will do this.

bin/c8.js
```diff
@@ -27,8 +27,9 @@ let argv = buildYargs().parse(instrumenterArgs)
     }
 
     process.env.NODE_V8_COVERAGE = argv.tempDirectory
-    foreground(hideInstrumenterArgs(argv), () => {
+    foreground(hideInstrumenterArgs(argv), done => {
       outputReport(argv)
+      done()
```
shinnn: Not needed. Even If no callback function is provided, foreground-child itself will do what you do here. https://github.com/tapjs/foreground-child/blob/v1.5.6/index.js#L22-L24

popomore: But c8 has pass ths cb argument without done.

shinnn: Indeed.

shinnn:
Again https://github.com/bcoe/c8/pull/82#pullrequestreview-231483541,

> Elaborate on what kind of problem this patch solves in https://github.com/bcoe/c8/pull/82#issue-274155174.

Otherwise LGTM.

bcoe:
@popomore (like @shinnn mentioned too) mind providing an example of the bug this fixes for you? it would be good to know so that we can write a test for it.

popomore:
Can you check this repo https://github.com/popomore/c8-mocha?

If you run `npm run cov`, the process will not exit. It seems that if ipc channel is opened, the listener should be cleaned as foreground-child done.

bcoe:
@popomore :+1: gotcha, thanks for clarifying; and simply ensuring that we call `done()` addresses the issue?

popomore:
Yes, foreground-child will remove the listener after calling done

popomore:
@shinnn Thank for your accepting

@bcoe when will be released? I want to test c8 in my code, comparing the performance with nyc.

bcoe:
@popomore please try `npm i c8@next`, I'll have the official release out a bit later this week.

popomore:
Thanks

PR #84, opened 2019-04-29
chore: test on Node.js v12.x

shinnn:
It would be better to test this tool with the next LTS version than the version whose life ends about one month later. https://github.com/nodejs/Release/tree/0e0b592273104d1cca9154588092654b932659b1#release-schedule

##### Checklist


test/fixtures/import.mjs
```diff
@@ -1,3 +1,3 @@
-import foo from './export'
+import foo from './export.mjs'
```
shinnn: In Node.js v12, `import` requires explicit file extensions unless `--es-module-specifier-resolution` flag is set to `node`.

bcoe:
@shinnn thank you for digging into this 👍

PR #85, opened 2019-04-29
feat: add support for source-maps

bcoe:
This adds initial support for source-maps, the brunt of the work was ultimately [this pull](https://github.com/istanbuljs/v8-to-istanbul/pull/19) in v8-to-istanbul. Coverage with source-maps seems to be working pretty well on my TypeScript projects (which makes me very happy 😄)

Screen Shot 2019-04-29 at 12 01 46 AM

## Some Notable Changes

* lines in `v8-to-istanbul` now start with a count of `1`, prior to this we were relying on the fact that V8 would always have an outer function that has an execution count of `1` for all its lines.
  * this was not working reliably, once mapping was applied, and I figured why not just default to `count = 1` if the outer function was a tautology any ways?
* a bunch of methods in `c8` are now async, this was because the heavy-lifting [source-map](https://github.com/mozilla/source-map) library is async.

## TODO

- [x] write integration tests for a variety of source-maps.
- [x] make sure this doesn't introduce any major performance regressions on the Node.js codebase, which uses c8's reporting functionality.

If you would like to experiment, I am publishing this to a `next` branch, as `4.0.0-candidate.0`.

BREAKING CHANGE: c8 will now load source-maps if possible and remap coverage accordingly

CC: @SimenB (please take this for a spin if interested), @hashseed (I think this unblocks me from some of the remapping work I was bumping into issues on), @sebmck (would love your extra set of eyes).

##### Checklist


bcoe:
@jdalton, I *finally* got source-maps implemented ... you might be able to play with `c8` a bit more in conjunction with ESM now?

jdalton:
So rad! Thank you 🎉 🌮 🎉

PR #88, opened 2019-05-02
feat: print warning and abort for non-compatible Node.js versions

bcoe:
rather than failing with a cryptic parsing error, or outputting no coverage, print a warning on older versions of Node.js.

##### Checklist

### DIFF
--- a/bin/c8.js
+++ b/bin/c8.js
@@ -1,35 +1,11 @@
 #!/usr/bin/env node
 'use strict'
 
-const fs = require('fs')
-const foreground = require('foreground-child')
-const { outputReport } = require('../lib/commands/report')
-const { promisify } = require('util')
-const rimraf = require('rimraf')
-const {
-  buildYargs,
-  hideInstrumenteeArgs,
-  hideInstrumenterArgs
-} = require('../lib/parse-args')
+const semver = require('semver')
 
-const instrumenterArgs = hideInstrumenteeArgs()
-let argv = buildYargs().parse(instrumenterArgs)
-
-;(async function run () {
-  if ([
-    'check-coverage', 'report'
-  ].indexOf(argv._[0]) !== -1) {
-    argv = buildYargs(true).parse(process.argv.slice(2))
-  } else {
-    if (argv.clean) {
-      await promisify(rimraf)(argv.tempDirectory)
-      await promisify(fs.mkdir)(argv.tempDirectory, { recursive: true })
-    }
-
-    process.env.NODE_V8_COVERAGE = argv.tempDirectory
-    foreground(hideInstrumenterArgs(argv), async (done) => {
-      await outputReport(argv)
-      done()
-    })
-  }
-})()
+if (semver.lt(process.version, '10.0.0')) {
+  /* c8 ignore next */
+  console.warn(`c8 requires Node v10.0.0 or greater, found ${process.version}`)
+} else {
+  require('./run')
+}
--- a/bin/c8.js
+++ b/bin/c8.js
@@ -1,8 +1,8 @@
 #!/usr/bin/env node
 'use strict'
 
-const minNode = `10.12.0`
 const semver = require('semver')
+const minNode = semver.minVersion(require('../package.json').engines.node)
 
 /* c8 ignore next 3 */
 if (semver.lt(process.version, minNode)) {
--- a/bin/c8.js
+++ b/bin/c8.js
@@ -1,11 +1,12 @@
 #!/usr/bin/env node
 'use strict'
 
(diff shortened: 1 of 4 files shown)


bin/c8.js
```diff
@@ -1,35 +1,11 @@
...
-    foreground(hideInstrumenterArgs(argv), async (done) => {
-      await outputReport(argv)
-      done()
-    })
-  }
-})()
+const semver = require('semver')
+
+/* c8 ignore next 3 */
+if (semver.lt(process.version, '10.12.0')) {
```
shinnn: I'd suggest:

```suggestion
if (semver.satisfies(process.version, require('../package.json').engines.node)) {
```

then we don't need to update two files whenever we change required Node.js version in the future.

bin/c8.js
```diff
@@ -1,35 +1,12 @@
...
-      await outputReport(argv)
-      done()
-    })
-  }
-})()
+const minNode = `10.12.0`
+const semver = require('semver')
+
+/* c8 ignore next 3 */
+if (semver.lt(process.version, minNode)) {
```
shinnn: I'd suggest:

```suggestion
if (semver.satisfies(process.version, require('../package.json').engines.node)) {
```

then we don't need to update two files whenever we change required Node.js version in the future.

bin/c8.js
```diff
@@ -1,35 +1,12 @@
...
+  console.warn(`c8 requires Node v${minNode} or greater, found ${process.version}`)
```
shinnn: Users expect `c8` to exit with non-zero code when they see this message.

bcoe: because coverage is an optional part of the testing process, I'd rather exit with `0`; this allows a user to run coverage for their entire testing matrix on a system like Travis:

`node 8`, `node 10, `node 11`, as an example, and they won't need special logic to prevent coverage from running on `node 8` (it just won't output anything) -- I actually have this use-case, which is why I added this feature.

bcoe: seem reasonable for now? would be happy to revisit in the future if this approach doesn't work well for people.

bcoe:
@shinnn hmm, this isn't quite the solution I was hoping for ... I guess we'd also need to have `foreground` child execute the original script ... I think you're right about exiting with `1`.

bin/c8.js
shinnn: OK, I agree on exiting with `0`. Thanks for the explanation.

> they won't need special logic to prevent coverage from running on `node 8` (it just won't output anything) 

Shouldn't the passed command, for example `tap` when they run `c8 tap`, be executed without coverage reporting, in this case?

bin/c8.js
```diff
@@ -1,35 +1,12 @@
...
-      await outputReport(argv)
-      done()
-    })
-  }
-})()
+const semver = require('semver')
+const minNode = semver.minVersion(require('../package.json').engines.node)
+
+/* c8 ignore next 3 */
+if (semver.lt(process.version, minNode)) {
```
shinnn: `semver.lt` doesn't work in this case because it requires the second argument to be an exact version, not a semver range.

```console
$ node -p "require('semver').satisfies(process.version, '>=10.2.0')" # it works
true

$ node -p "require('semver').lt(process.version, '>=10.2.0')" # it doesn't
[eval]:1
require('semver').satisfiles('12.1.0', '>=10.2.0')
                  ^

TypeError: require(...).satisfiles is not a function
    at [eval]:1:19
    at Script.runInThisContext (vm.js:123:20)
    at Object.runInThisContext (vm.js:313:38)
    at Object.<anonymous> ([eval]-wrapper:9:26)
    at Module._compile (internal/modules/cjs/loader.js:759:30)
    at evalScript (internal/process/execution.js:79:25)
    at internal/main/eval_string.js:23:3
```

bin/c8.js
shinnn: (@bcoe already posted https://github.com/bcoe/c8/pull/88#issuecomment-488570240 while I'm writing https://github.com/bcoe/c8/pull/88#discussion_r280305528)

bin/run.js
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env node
```
shinnn:
```suggestion
```

This file doesn't need to have a shebang.

bcoe:
I don't think this PR adds enough value, vs., communicating to the user the minimum engine in the `package.json`.

bin/c8.js
shinnn: My fault. https://github.com/bcoe/c8/pull/88/files is wrong.

PR #91, opened 2019-05-03
fix: remove the unmaintained mkdirp dependency

shinnn:
https://github.com/bcoe/c8/commit/206b83f156ff93d1a7b211d61bd4fd8e9a4cd95f introduced the `mkdirp` dependency to `c8`, but I'm very opposed to use such a no longer maintained package.

For feature detection, we can check the existence of [`fs.promises`](https://nodejs.org/dist/latest-v12.x/docs/api/fs.html#fs_fs_promises_api) which is only available in Node.js >= 10.0.0.

##### Checklist

### DIFF
--- a/bin/c8.js
+++ b/bin/c8.js
@@ -3,7 +3,7 @@
 
 const foreground = require('foreground-child')
 const { outputReport } = require('../lib/commands/report')
-const mkdirp = require('mkdirp')
+const { promises } = require('fs')
 const { promisify } = require('util')
 const rimraf = require('rimraf')
 const {
@@ -25,7 +25,9 @@ async function run () {
       await promisify(rimraf)(argv.tempDirectory)
     }
     // allow c8 to run on Node 8 (coverage just won't work).
-    await promisify(mkdirp)(argv.tempDirectory)
+    if (promises) {
+      await promises.mkdir(argv.tempDirectory, { recursive: true })
+    }
 
     process.env.NODE_V8_COVERAGE = argv.tempDirectory
     foreground(hideInstrumenterArgs(argv), async (done) => {
--- a/bin/c8.js
+++ b/bin/c8.js
@@ -21,7 +21,8 @@ async function run () {
   ].indexOf(argv._[0]) !== -1) {
     argv = buildYargs(true).parse(process.argv.slice(2))
   } else {
-    // allow c8 to run on Node 8 (coverage just won't work).
+    // fs.promises was not added until Node.js v10.0.0, if it doesn't
+    // exist, assume we're Node.js v8.x and skip coverage.
     if (!promises) {
       foreground(hideInstrumenterArgs(argv))
       return
--- a/bin/c8.js
+++ b/bin/c8.js
@@ -21,14 +21,17 @@ async function run () {
   ].indexOf(argv._[0]) !== -1) {
     argv = buildYargs(true).parse(process.argv.slice(2))
   } else {
+    // allow c8 to run on Node 8 (coverage just won't work).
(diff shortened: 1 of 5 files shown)


bcoe:
@shinnn the problem isn't the lack of promisify, it's the fact that `recursive` doesn't work on older Node.js versions, leading to this bug:

```bash
Benjamins-MBP-2:c8 benjamincoe$ ./bin/c8.js node foo.js
(node:15352) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 2): Error: ENOENT: no such file or directory, scandir './coverage/tmp'
(node:15352) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

Perhaps you could either:

* check to see if the directory actually exists, and exit with `0` and warn if it doesn't after attempting to create it.
* or switch to using [make-dir](https://www.npmjs.com/package/make-dir) which is maintained, and falls back to `recursive` if it's available.

shinnn:
Actually 491aebb81d87e590466206caebd45122c11bc14e doesn't work properly on Node.js 8, but

> it's the fact that `recursive` doesn't work on older Node.js versions

is not correct. The real problem is that `c8` still tries to output coverage reports even though `NODE_V8_COVERAGE` is not supported. Check the stack trace.

```
(node:82468) UnhandledPromiseRejectionWarning: Error: ENOENT: no such file or directory, scandir './coverage/tmp'
    at fs.readdirSync (fs.js:904:18)
    at Report._loadReports (/Users/shinnn/c8/lib/report.js:124:19)
```

shinnn:
I added a new commit 1f938097f340a5a42cc39c91c2dc60a926ba4a85 which suppresses the error mentioned in https://github.com/bcoe/c8/pull/91#issuecomment-489354330.

@bcoe Can you pull the up-to-date branch and try it again with Node.js 8?

bin/c8.js
```diff
@@ -21,12 +21,17 @@ async function run () {
   ].indexOf(argv._[0]) !== -1) {
     argv = buildYargs(true).parse(process.argv.slice(2))
   } else {
+    // allow c8 to run on Node 8 (coverage just won't work).
+    if (!promises) {
```
bcoe: maybe just pull in `semver` a dependency and:

```
if (semver.lt('10.12.0', process.version)) {
  console.warn('native test coverage was not added to V8 until 10.12.0')
  foreground(hideInstrumenterArgs(argv))
  return
}
```

that way we're not relying on `util.promisify` not being back-ported.

👍 thanks for your work on this; I'll be back a bit later today.

shinnn: > that way we're not relying on `util.promisify` not being back-ported.

To be accurate, the API used for feature detection here is not `util.promisify` but [`fs.promises`](https://nodejs.org/api/fs.html#fs_fs_promises_api).

I can hardly imagine the Node.js team decides to back-port `fs.promises` to [the maintenance LTS version](https://github.com/nodejs/Release/tree/87c98116e471e86a66dc39ebe703d3394323f8da#release-schedule).

bcoe: perhaps just add a comment then:

```
// fs.promises was not added until Node v10.0.0, if it doesn't
// exist, assume we're Node v8.x and skip coverage.
```

shinnn: 938a46391105b4f56bc1c07144dba362da74b4b1

bcoe:
@shinnn thank you!

shinnn:
@bcoe No problem. Thanks for the review.

PR #152, opened 2019-10-24
feat!: use Node.js' source-map cache, to support tools like ts-node

bcoe:
BREAKING CHANGE: Node.js' source-map and lineLength cache is now used  to remap coverage output (this allows tools like ts-node to be supported, which transpile at runtime).


bcoe:
see: https://github.com/nodejs/node/issues/29865

PR #168, opened 2019-11-28
feat(istanbul-reports)!: lcov reports now use relative paths

bcoe:
The next major release of `nyc` uses updated Istanbul APIs, we should move to them too.

The feature I'm most excited about is support for relative paths in lcov reports (which is required for merging reports).

CC: @coreyfarrell, @j03m


coreyfarrell:
> This looks good to me. I trust the alpha is stable enough?

The istanbul stuff is inches away from final, I don't expect any significant changes.  Honestly at this point not sure why I labeled the releases alpha instead of beta.

PR #192, opened 2020-01-25
fix: stop duplicating names on case-insensitive file systems

j03m:
I found that on win32, if the v8 script name was set using "legal" (I'm using that term loosely, because it probably shouldn't be done by the embedder), but mismatched casing for a file path, it would result in `--all` reporting duplicate records for an entry. 

For example where node's `path.resolve` might provide you with `c:\src\main.js` it would be perfectly fine on windows to tell v8 that the script resource was `c:\src\MAIN.js`. If you then try to measure coverage with all, the mismatch in casing between v8's blobs and node's file system results cause the system to report duplicate empty records for files that are actually in the blobs, just differ on case.

Prior to my fix test output vs the correct snapshot  would like like:

```
     AssertionError: expected value to match snapshot c8 --all should, given path casing mismatches between path.resolve and v8 output, NOT contain duplicate report entries 1
      + expected - actual

       ",--------------|---------|----------|---------|---------|-------------------
       File          | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
       --------------|---------|----------|---------|---------|-------------------
      -All files     |   35.29 |    54.55 |      25 |   35.29 |                   
      - vanilla      |   39.13 |       60 |   33.33 |   39.13 |                   
      +All files     |   64.29 |    66.67 |      50 |   64.29 |                   
      + vanilla      |   78.26 |       75 |     100 |   78.26 |                   
         LOADED.js   |   73.68 |    71.43 |     100 |   73.68 | 4,5,16-18
         MAIN.js     |     100 |      100 |     100 |     100 |
      -  loaded.js   |       0 |        0 |       0 |       0 | 1-19              
      -  main.js     |       0 |        0 |       0 |       0 | 1-4               
        vanilla/dir  |       0 |        0 |       0 |       0 |
         unloaded.js |       0 |        0 |       0 |       0 | 1-5
       --------------|---------|----------|---------|---------|-------------------
       ,"

      at Context.<anonymous> (test\integration.js:457:40)
      at processImmediate (internal/timers.js:439:21)
```

##### Checklist
- [ x ] `npm test`, tests passing
- [ x ] `npm run test:snap` (to update the snapshot)
- [ x ] tests and/or benchmarks are included

### DIFF
--- a/lib/path-set.js
+++ b/lib/path-set.js
@@ -1,0 +1,25 @@
+/**
+ * A unique set of paths. Case insensitive on win32
+ */
+class PathSet extends Set {
+  constructor () {
+    super()
+    this._win32 = process.platform === 'win32'
+  }
+
+  add (value) {
+    if (this._win32) {
+      value = value.toLowerCase()
+    }
+    super.add(value)
+  }
+
+  has (value) {
+    if (this._win32) {
+      value = value.toLowerCase()
+    }
+    return super.has(value)
+  }
+}
+
+module.exports = PathSet
--- a/lib/path-set.js
+++ b/lib/path-set.js
@@ -12,6 +12,7 @@ class PathSet extends Set {
       value = value.toLowerCase()
     }
     super.add(value)
+    return this
   }
 
   has (value) {
--- a/lib/report.js
+++ b/lib/report.js
@@ -3,6 +3,7 @@ const furi = require('furi')
 const libCoverage = require('istanbul-lib-coverage')
 const libReport = require('istanbul-lib-report')
 const reports = require('istanbul-reports')
+const PathSet = require('./path-set')
 const { readdirSync, readFileSync, statSync } = require('fs')
 const { isAbsolute, resolve, extname } = require('path')
 const getSourceMapFromFile = require('./source-map-from-file')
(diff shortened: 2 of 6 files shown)


lib/path-set.js
```diff
@@ -0,0 +1,26 @@
...
+ */
+class PathSet extends Set {
+  constructor () {
+    super()
+    this._win32 = process.platform === 'win32'
+  }
+
+  add (value) {
+    if (this._win32) {
+      value = value.toLowerCase()
```
bcoe: My concern is that I think Windows can have case sensitivity enabled? [see this article](https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2008-R2-and-2008/cc725747(v=ws.11)?redirectedfrom=MSDN).

I wonder if we could find a way to feature detect case sensitivity, rather than basing it on Windows?

j03m: Oh I didn't know that. Will review. Thanks!

SimenB: I looked into that for jest and didn't find a nice way (suggested solutions was writing a file and reading it with different casing, but it had some caveats).

If you _do_ find a nice way, please publish it as a separate module and we can use it in Jest as well 😀

bcoe: @SimenB did you publish a module for the approach used by Jest? otherwise I'd be happy to pull together a module that at least exposes the hack.

j03m:
Just noting an update here - we're going to put this on hold in favor of creating a separate module in istanbul that can detect case sensitivity of the file system. See: https://github.com/istanbuljs/test-exclude/pull/44

lib/path-set.js
SimenB: No, we gave up and just check for `platform === 'win32'` 😞 So please put together a module! Would be cool if Node itself could expose something...

j03m:
Closing this. Will try to create a module this weekend.

PR #198, opened 2020-02-07
feat: adds TypeScript definitions

Toxicable:
fixes 195

##### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,8 @@
   "name": "c8",
   "version": "7.0.1",
   "description": "output coverage reports using Node.js' built in coverage",
-  "main": "index.js",
+  "main": "./index.js",
+  "types": "./index.d.ts",
   "bin": "./bin/c8.js",
   "repository": {
     "type": "git",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -33,6 +34,7 @@
   "dependencies": {
     "@bcoe/v8-coverage": "^0.2.3",
     "@istanbuljs/schema": "^0.1.2",
+    "@types/istanbul-lib-report": "^3.0.0",
```
bcoe: How complex is the watermark type, I wonder if we could get away with not pulling in the `@types/istanbul-lib-report` dependency, and just including the description inline.

Toxicable: thats fair, inlined and removed the dep

Toxicable:
Tested downstream here https://github.com/bazelbuild/rules_nodejs/pull/1135/files#diff-cf68df009e956fdc1b73ff16773264f1R80
The CI in that PR will fail (as expected) but that TS file compiles which indicates it's working as expected

Toxicable:
It was tested agains tthat commit I just pushed ^

Issue #149, opened 2019-09-12
Running in node 12 generates an invalid XML when running using '--reporter cobertura'


assaf-xm:
Version: v12.9.1
Platform: 4.4.0-130-generic

We use c8 with the cobertura reporter.
Moving from node 10 to node 12 the generated XML was invalid and crashed the Jenkins Cobertura  publisher.

It seems that node 12 the '(anonymous function)' now becomes '`<computed>`' and the '<' character is invalid inside an attribute (in the cobertura-coverage.xml).

For example:
Node12
`<method name="self.cam.<computed>.handler" hits="6" signature="()V">`

Node10
`<method name="self.cam.(anonymous function).handler" hits="6" signature="()V">`

The Jenkins error:
FATAL: Unable to parse .../builds/3622/coverage.xml
09:17:48 hudson.util.IOException2: Cannot parse coverage results
09:17:48 	at hudson.plugins.cobertura.CoberturaCoverageParser.parse(CoberturaCoverageParser.java:84)
09:17:48 	at hudson.plugins.cobertura.CoberturaCoverageParser.parse(CoberturaCoverageParser.java:52)
09:17:48 	at hudson.plugins.cobertura.CoberturaPublisher.perform(CoberturaPublisher.java:593)
09:17:48 	at hudson.tasks.BuildStepCompatibilityLayer.perform(BuildStepCompatibilityLayer.java:79)
09:17:48 	at hudson.tasks.BuildStepMonitor$1.perform(BuildStepMonitor.java:20)
09:17:48 	at hudson.model.AbstractBuild$AbstractBuildExecution.perform(AbstractBuild.java:741)
09:17:48 	at hudson.model.AbstractBuild$AbstractBuildExecution.performAllBuildSteps(AbstractBuild.java:690)
09:17:48 	at hudson.model.Build$BuildExecution.post2(Build.java:186)
09:17:48 	at hudson.model.AbstractBuild$AbstractBuildExecution.post(AbstractBuild.java:635)
09:17:48 	at hudson.model.Run.execute(Run.java:1840)
09:17:48 	at hudson.model.FreeStyleBuild.run(FreeStyleBuild.java:43)
09:17:48 	at hudson.model.ResourceController.execute(ResourceController.java:97)
09:17:48 	at hudson.model.Executor.run(Executor.java:429)
[...]

PR #203, opened 2020-03-08
fix: escaping issue with cobertura reporter

bcoe:
See: https://github.com/istanbuljs/istanbuljs/pull/534

fixes: #149


j03m:
Looks sound, but looking at the snapshot diff, I think in the integration test you may need to scrub the "source" tag as well?

PR #205, opened 2020-03-18
feat: Allow --all to support multiple directories from Report

j03m:
****DRAFT: 

We're currently leveraging the `index.js` programmatically via `require("c8")` for a build system that is bit non standard in that multiple `dist` directories for various projects can be leveraged in a workspace that is outside the cwd of where the test are run. 

This change modifies the interface to `Report` but not the CLI in the following ways:

* It allows `src` to be supplied as a Array of string paths. 
* If `src` isn't supplied, src is overriden to the default `cwd`, the main CLI doesn't currently allow these directories to be supplied to avoid over complicating the CLI interface.
* An additional flag for `allowExternal` is supplied and passed to `test-exlude` to indicate that not all paths will be relative. Not supplying this flag causes `test-exclude` to reject anything outside of `cwd`.
 
Please don't merge this just yet as I'm still testing, writing tests. But I wanted to get your thoughts if this was a feasible enhancement.

This yea

##### Checklist

- [ x ] `npm test`, tests passing

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -17,7 +17,8 @@ export declare class Report {
         wrapperLength?: number,
         resolve?: string,
         all?: boolean,
+        src?: Array<string>,
+        allowExternal: boolean
     })
-
     run(): Promise<void>;
 }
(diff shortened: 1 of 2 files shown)


j03m:
Closing to re-open as draft. *doh*

Issue #216, opened 2020-05-02
Unit tests failing for clean clone of master


j03m:
* **Version**: Hash : 56b77232ddc701f3334a86a5bd88bb4bbbb2d0e0
* **Platform**: osx

Seems like something went off the rails with a previous merge. 

**Edit to add: Seems like pulling in the new reporter broke the snapshots. Previous examples that were in the format of `\d+,\d+` have turned into `\d-\d`. If we're okay with that I will submit a PR to update the snapshots and fix whatever is up with the cobertura test

Running `npm test` results in: 

```

> c8@7.1.1 test /Users/jmordetsky/c8
> node ./bin/c8.js mocha --timeout=8000 ./test/*.js



  c8
    ✓ reports coverage for script that exits normally (805ms)
    ✓ supports exeternally set NODE_V8_COVERAGE (822ms)
    ✓ merges reports from subprocesses together (429ms)
    ✓ allows relative files to be included (477ms)
    ✓ exits with 1 when report output fails (520ms)
    1) cobertura report escapes special characters
    check-coverage
      ✓ exits with 0 if coverage within threshold (182ms)
      ✓ exits with 1 if coverage is below threshold (185ms)
      ✓ allows threshold to be applied on per-file basis (184ms)
      ✓ allows --check-coverage when executing script (803ms)
    report
      ✓ generates report from existing temporary files (224ms)
      ✓ supports --check-coverage, when generating reports (231ms)
    ESM Modules
      2) collects coverage for ESM modules
    /* c8 ignore next */
      3) ignores lines with special comment
    source-maps
      TypeScript
        4) remaps branches
        5) remaps classes
      UglifyJS
        6) remaps branches
        7) remaps classes
      nyc
        ✓ remaps branches (293ms)
        8) remaps classes
    ts-node
      9) reads source-map from cache, and applies to coverage
    --all
      10) reports coverage for unloaded js files as 0 for line, branch and function
      11) reports coverage for unloaded transpiled ts files as 0 for line, branch and function
      12) reports coverage for unloaded ts files as 0 for line, branch and function when using ts-node
      13) should allow for --all to be used in conjunction with --check-coverage
      ✓ should allow for --all to be used with the check-coverage command (2 invocations) (832ms)

  c8 on Node.js < 10
    ✓ skip coverage (169ms)

  parse-args
    hideInstrumenteeArgs
      ✓ hides arguments passed to instrumented app
    hideInstrumenterArgs
[...]

PR #217, opened 2020-05-02
chore: fix integration test snapshot

j03m:
Fixes https://github.com/bcoe/c8/issues/216

It looks like with the reporter update and the cobertura update somehow the snapshots managed to get out of date. I just ran `npm run test:snap` to update and manually verified there wasn't a serious regression.

##### Checklist


test/integration.js.snap
```diff
@@ -8,7 +8,7 @@ File               | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
 -------------------|---------|----------|---------|---------|-------------------
 All files          |   86.21 |    91.67 |   66.67 |   86.21 |                   
  async.js          |     100 |      100 |     100 |     100 |                   
- c8-ignore-next.js |   90.91 |      100 |     100 |   90.91 | 21-22
```
j03m, line 11: Outside of cobertura the only regressions are the change from `\d+-\d+` to `\d+,\d+` which seems like the *correct* value for 1 line ranges?

j03m:
Ignore, I didn't update my deps. :/

PR #228, opened 2020-06-19
chore: simplify config loader

lukeed:
Hey there, thank you for `c8` :)

I was trying to figure out how it works recently & noticed you're loading 7 dependencies to hunt for a config file.
While it's not a _significant_ part of your dependency chain, it's still a [branch that can be removed](https://npm.anvaka.com/#!/view/2d/c8).

Instead, `escalade` is 21 lines of code and has no dependencies. It's also [faster than `find-up`](https://github.com/lukeed/escalade#benchmarks) and supports native ESM when `c8` moves its source that way.

There's also a security advantage – `find-up` will continue traversal into your `/` directory (wtf) before giving up. Escalade stops at the `process.cwd()` directory, which one would expect...

FWIW, I may have just _released_ `"escalade"` but I've been using this internally for a while.
Didn't set out to replace `find-up`, but once these comparison points dawned on me, I figured it needs sharing.

##### Checklist

### DIFF
--- a/lib/parse-args.js
+++ b/lib/parse-args.js
@@ -1,11 +1,12 @@
 const defaultExclude = require('@istanbuljs/schema/default-exclude')
-const findUp = require('find-up')
+const escalade = require('escalade/sync')
 const { readFileSync } = require('fs')
 const Yargs = require('yargs/yargs')
 const parser = require('yargs-parser')
 const { resolve } = require('path')
 
-const configPath = findUp.sync(['.c8rc', '.c8rc.json', '.nycrc', '.nycrc.json'])
+const configFiles = ['.c8rc', '.c8rc.json', '.nycrc', '.nycrc.json']
+const configPath = escalade('.', (dir, names) => configFiles.find(x => names.includes(x)))
 const config = configPath ? JSON.parse(readFileSync(configPath)) : {}
 
 function buildYargs (withCommands = false) {
--- a/package-lock.json
+++ b/package-lock.json
@@ -553,6 +553,11 @@
         "is-symbol": "^1.0.2"
       }
     },
+    "escalade": {
+      "version": "1.0.0",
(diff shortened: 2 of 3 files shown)


lukeed:
I'm not sure what's changed since two days ago, but a fresh `master` clone also fails for me on Node 10.x

bcoe:
Hey @lukeed I just tested locally with `v10.20.1`, and tests are passing. Are you seeing errors locally as well?

I appreciate the patch, but I'm a little hesitant to swap out `find-up`, mainly because it has a few 1000s dependents, vs., `escalade`, which hasn't been adopted by many projects yet (it looks like).

Once it's been adopted by a few projects, I'd happily revisit the swap (sounds like it does have some benefits) ... I mainly don't want to be a guinea pig in c8, this library already has a lot of moving parts, and I have limited time to work on it.

lukeed:
Sure, understandable :) It's just a new repo, but not new code – I pulled it from existing projects.

Locally I tested with 10.13.0, and your CI is running 10.18 – so maybe 10.20 is carrying something special?

I'm happy to revisit this in the future. I'll keep the branch handy

bcoe:
@lukeed I'm probably sometimes overly hesitant in pulling in deps, I think in this case I'd love to see the library used in just a few projects fir a bit first, even if it's just a few of yours.

lukeed:
Nope, I understand completely! I share the same approach. 

I would have pulled it into some of my own modules first, but all of my (public) CLI apps are strict/opinionated about where the config files, if any, should live. I have a monorepo-oriented CLI coming up, and that's the first (public) use case that actually requires some config guesswork/fs prodding.

bcoe:
@lukeed happy to revisit this when we consider the move to ESM, something which I am interested in in the future.

lukeed:
No problem :) Btw, Browserslist has been using `escalade` for about two weeks now

Issue #271, opened 2020-12-08
support `c8 ignore start`, `c8 ignore stop`


connorjclark:
Since ignore AST nodes isn't an option, ignoring functions becomes difficult:

```
/* c8 ignore next 11 */
function wrapRuntimeEvalErrorInBrowser(err) {
  err = err || new Error();
  const fallbackMessage = typeof err === 'string' ? err : 'unknown error';

  return {
    __failedInBrowser: true,
    name: err.name || 'Error',
    message: err.message || fallbackMessage,
    stack: err.stack || (new Error()).stack,
  };
}
```

Most changes to this function will require modifying the ignore comment.

With istanbul it's trivial: 

```
/* istanbul ignore next */
function wrapRuntimeEvalErrorInBrowser(err) {
  err = err || new Error();
  const fallbackMessage = typeof err === 'string' ? err : 'unknown error';

  return {
    __failedInBrowser: true,
    name: err.name || 'Error',
    message: err.message || fallbackMessage,
    stack: err.stack || (new Error()).stack,
  };
}
```

But no AST. instead, what about a "start ignoring" & "stop ignoring" comment?

bcoe:
@connorjclark this seems like a reasonable compromise on not being able to ignore specific Node types.

I've been hesitant to want to introduce a JavaScript parser to the codebase.

Issue #272, opened 2020-12-09
snapshot tests fail on fresh checkout


connorjclark:
A fresh checkout results in this after running `npm run test:snap`

![image](https://user-images.githubusercontent.com/4071474/101697505-914b9d80-3a3d-11eb-8fe1-e0f970caa5bb.png)

Node: v12.13.0

bcoe:
@connorjclark what Node.js version are you running, I just deleted `pacakge-lock.json` and `node_modules`, and ran:

`npm run test`

And the snapshots don't throw any errors:

`41 passing (16s)`

My environment:

```
bencoe-macbookpro3:c8 bencoe$ node --version
v15.0.1
```

connorjclark:
Node 12.

I missed this line in CONTRIBUTING:

> Make sure you have installed the latest version of Node.js

Tests pass for me using Node 15.

PR #273, opened 2020-12-09
feat: support ignore start/stop comment

connorjclark:
fixes bcoe/c8#271

blocked on https://github.com/istanbuljs/v8-to-istanbul/pull/130

Note that the first commit is simply rebaselining the failing snapshot tests I get on my machine (see #272). See the second commit to see only what this PR really changed.


bcoe:
@connorjclark thank you for this contribution, and sorry for the slow review.

Could I bother you to explicitly update the `v8-to-istanbul` dependency to `7.1.0`? which should be published shortly with your additions.

connorjclark:
Using Node 15 I was able to undo the test "updates" in commit 1.

I also noticed the package-lock went from `lockfileVersion: 1` to `lockfileVersion: 2` when I ran `npm i`, using node 15 (`nvm 15` gives me npm `7.0.15`). I didn't commit those changes.

connorjclark:
I'm confused... apparently the tests expect Node 15 (Node 12 fails for me), but CI is running tests and expects Node 10, 12, and 14 to pass?

bcoe:
@connorjclark I'm not 100% sure what was going on here.

Node 10 snapshots need to be taken separately from other Node versions, because Node 10 is missing a lot of functionality -- I pushed Noe 10 snapshots taken from my computer and everything worked.

---

Testing can be a bit finicky because I've gradually been patching bugs in v8, so different Node.js versions will have slightly different coverage counts.

test/integration.js
```diff
@@ -294,6 +294,20 @@ describe('c8', () => {
     })
   })
 
+  describe('/* c8 ignore start/stop */', () => {
+    it('ignores lines with special comment', () => {
+      const { output } = spawnSync(nodePath, [
+        c8Path,
+        '--exclude="test/*.js"',
+        '--clean=false',
+        '--temp-directory=tmp/normal',
```
bcoe, line 303: @connorjclark part of the problem was here, we should use a different tmp directory for each test (it was an oversight that we were already reusing `tmp/normal`).

I believe this was contributing to nondeterministic behavior.

bcoe (approved):

bcoe:
@connorjclark okay, this is released in `c8@7.4.0` apologies for the slow turnaround.

PR #283, opened 2021-02-01
feat(all): handle base64 inline source maps

### LINKED ISSUE #280: Report not able to handle base64 files when using all: true
Reports aren't able to read base64 source maps files
Looking at this line here: https://github.com/bcoe/c8/blob/master/lib/report.js#L189
It's not able to read files which have their content encoded as a base64 path.
Example: `path/to/file:application/json;base64,{base64 content}`

Right now it'll throw an error like:
> ENAMETOOLONG: name too long, open path/to/file:application/json;base64,{base64 content}

bcoe:
Pulls in Source Map handling logic from Node.js

@Toxicable could you please give this a try in your testing repositories, and with bazel?

Fixes #280


Toxicable:
Looks like that did the trick, thanks!

PR #301, opened 2021-04-27
fix: address bugs with source remapping on Windows

### LINKED ISSUE #300: V8 source map cache is not used on Windows
* **Version**: v16.0.0, v15.14.0
* **Platform**: Windows 10 64-bit

## Summary

Somewhere along the node v15 line, V8 coverage data output by node changed format on Windows. Specifically, file URLs within the `source-map-cache` changed from

```
file://C:\some\path\main.js
```
to
```
file:///C:/some/path/main.js
```

(Note the extra forward slash in front)

I haven't bisected further but v15.0.0 is using backslashes and v15.14.0 is using forward slashes.

This makes `_getSourceMap` unable to correctly get cached source maps (since it searches using backslashes but keys in the `sourceMapCache` use forward slashes).

## Symptoms

This directly affects use cases where the source maps are virtual (e.g. `ts-node`), causing c8 to report incorrect locations for uncovered code. Worse-case scenario is when the reported range is outside of the original source size (e.g. position 1000 in generated code is uncovered but original source is only 500 bytes), in which case the uncovered branch/function/line is silently discarded.

This also affects cases where source maps are not virtual (i.e. exist in actual fs), but in an unexpected way. v8-to-instanbul is also affected by a [similar bug](https://github.com/istanbuljs/v8-to-istanbul/issues/143), and would throw when given a source map with file URL as source paths. However, since c8 cannot find the source map, no source maps are passed to v8-to-instanbul, avoiding the buggy code path.

## Reproduction

Run the following file with c8 and ts-node

```typescript
export default () => {
	console.log('uncovered')
}

// Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
```

Node v14.16.1 & v15.0.0 result:

image

Node v15.14.0 & v16.0.0 result:

image

## Fix

This can be fixed by normalizing file URLs in the `source-map-cache` to (and always querying the `sourceMapCache` using) forward slashes, with `url.fileURLToPath` and `url.pathToFileURL`.

This, however, breaks some test due to the v8-to-instanbul bug noted above.

clemyan:
##### Checklist

Fixes #300 with the fix outline in the issue.

As mentioned in the issue, fixing it reveals an upstream bug that causes test to fail. I have verified that on both node v14.16.1 and v16.0.0:

- Tests pass on Linux (docker)
- Locally linking to a fixed version of `v8-to-istanbul` makes tests pass on Windows

### DIFF
--- a/lib/report.js
+++ b/lib/report.js
@@ -1,10 +1,10 @@
 const Exclude = require('test-exclude')
-const furi = require('furi')
 const libCoverage = require('istanbul-lib-coverage')
 const libReport = require('istanbul-lib-report')
 const reports = require('istanbul-reports')
 const { readdirSync, readFileSync, statSync } = require('fs')
 const { isAbsolute, resolve, extname } = require('path')
+const { pathToFileURL, fileURLToPath } = require('url')
 const getSourceMapFromFile = require('./source-map-from-file')
 // TODO: switch back to @c88/v8-coverage once patch is landed.
 const v8toIstanbul = require('v8-to-istanbul')
@@ -139,8 +139,8 @@ class Report {
    */
   _getSourceMap (v8ScriptCov) {
     const sources = {}
-    if (this.sourceMapCache[`file://${v8ScriptCov.url}`]) {
-      const sourceMapAndLineLengths = this.sourceMapCache[`file://${v8ScriptCov.url}`]
+    const sourceMapAndLineLengths = this.sourceMapCache[pathToFileURL(v8ScriptCov.url).href]
+    if (sourceMapAndLineLengths) {
       // See: https://github.com/nodejs/node/pull/34305
       if (!sourceMapAndLineLengths.data) return
       sources.sourceMap = {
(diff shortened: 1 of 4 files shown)


lib/report.js
```diff
@@ -1,10 +1,10 @@
 const Exclude = require('test-exclude')
-const furi = require('furi')
```
bcoe, line 2: can we remove this from the `package.json` as well?

bcoe (changes requested):
Thank you for the contribution 👏 

One nit, let's remove `furi` from `package.json`?

bcoe:
@clemyan also looks like a slight hiccup with Node 10, it might just be you need to run `test:snap` on that version?

lib/report.js
clemyan: We can indeed

clemyan:
Well I wasn't sure if updating snapshots was the right move since some tests fails on Windows. I will update the snapshots with a Linux docker container in a few hours.

bcoe (approved):

bcoe:
@clemyan thank you for digging so deeply into this problem 👍 great contribution.

PR #331, opened 2021-09-30
feat: add --extension option

DylanPiercey:
Adds the `extension` option from nyc.
I've setup the CLI to support the same arguments as nyc (which translates to the config file also).

Currently without this change there is no way to use c8 with require hooks for tools without the default extensions (https://github.com/istanbuljs/schema/blob/master/default-extension.js). For example using a Marko, Svelte, Vue, etc require hook will not work.

With this PR you can currently set it to `false` to disable this extension check, same as nyc. It may make sense to use different defaults than nyc here, and just allow any extension. I've currently opted to just keep a similar api to nyc.

##### Checklist


bcoe:
@DylanPiercey have you tested with these changes. `c8` itself doesn't perform instrumentation, like `nyc`, rather it just sets an environment variable which is consumed by Node.js. So, if I'm understanding correctly, c8 is working for these alternate extensions, but then they get dropped by `test-exclude`?

DylanPiercey:
@bcoe I did test the changes and it worked. Did not test with node 10 though.

> they get dropped by test-exclude

Yeup, that is what was happening.

bcoe (approved):

bcoe:
@DylanPiercey thank you kindly for the contribution, sorry this took so long to merge.

PR #332, opened 2021-10-01
feat: `--100`

### LINKED ISSUE #329: support shorthand for 100% test coverage check
* **Version**: v16.9.1
* **Platform**: Mac OS X 10.15.7 (19H1417)

I really like [`tap`'s `--100` flag](https://node-tap.org/docs/coverage/100/). Could we add something similar to `c8`? It would turn

```
c8 --check-coverage --lines 100 --functions 100 --branches 100 --statements 100  node test.js
```

into 

```
c8 --100 node test.js
```

gr2m:
closes #329

### DIFF
--- a/lib/parse-args.js
+++ b/lib/parse-args.js
@@ -107,6 +107,12 @@ function buildYargs (withCommands = false) {
       description: 'check thresholds per file',
       type: 'boolean'
     })
+    .option('100', {
+      default: false,
+      group: 'Coverage thresholds',
+      description: 'shortcut for --check-coverage --lines 100 --functions 100 --branches 100 --statements 100',
+      type: 'boolean'
+    })
     .option('temp-directory', {
       describe: 'directory V8 coverage data is written to and read from',
       default: process.env.NODE_V8_COVERAGE
@@ -144,6 +150,18 @@ function buildYargs (withCommands = false) {
       return true
     })
     .epilog('visit https://git.io/vHysA for list of available reporters')
+
+  yargs.middleware((argv) => {
+    if (!argv[100]) return argv
+
+    return {
+      ...argv,
+      branches: 100,
+      functions: 100,
+      lines: 100,
+      statements: 100,
+    }
+  })
 
   const checkCoverage = require('./commands/check-coverage')
   const report = require('./commands/report')
--- a/lib/parse-args.js
+++ b/lib/parse-args.js
@@ -151,17 +151,18 @@ function buildYargs (withCommands = false) {
     })
     .epilog('visit https://git.io/vHysA for list of available reporters')
 
-  yargs.middleware((argv) => {
-    if (!argv[100]) return argv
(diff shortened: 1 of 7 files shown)


lib/parse-args.js
```diff
@@ -145,6 +151,18 @@ function buildYargs (withCommands = false) {
...
+    if (!argv[100]) return argv
+
+    return {
+      ...argv,
+      branches: 100,
+      functions: 100,
+      lines: 100,
+      statements: 100,
+    }
+  })
```
gr2m, line 164: this code doesn't seem to be executed. I haven't worked with yargs middleware before but it seems like a good approach here? Do you know why the code is not executed?

test/integration.js
```diff
@@ -211,6 +211,23 @@ describe('c8', () => {
...
+      const { output, status } = spawnSync(nodePath, [
+        c8Path,
+        '--exclude="test/*.js"',
+        '--temp-directory=tmp/check-coverage',
+        '--100',
+        nodePath,
+        require.resolve('./fixtures/normal')
+      ])
+      console.log(`output`);
+      console.log(output.toString("utf8"));
```
gr2m, line 225: reminder to self: remove once things work as they should

lib/parse-args.js
bcoe: This seems like a reasonable approach to me, and seems like it should work, perhaps:

```

yargs.middleware((argv) => {
    if (!argv['100']) return argv

    return {
      ...argv,
      branches: 100,
      functions: 100,
      lines: 100,
      statements: 100,
    }
}
```

I wouldn't expect the argument key itself to be coerced into an integer.

lib/parse-args.js
```diff
@@ -145,6 +151,18 @@ function buildYargs (withCommands = false) {
     })
     .epilog('visit https://git.io/vHysA for list of available reporters')
 
+  yargs.middleware((argv) => {
+    if (!argv[100]) return argv
```
bcoe, line 155:
```suggestion
    if (!argv['100']) return argv
```

bcoe (approved):
This general approach looks good to me, let me know when you're ready for review 👍

lib/parse-args.js
gr2m: my problem here is that the middleware is not applied at all. I think I'm missing something obvious. Any idea what it could be?

bcoe: I tested locally, it seems like this is a bug that I fixed in yargs@17, at least it works as expected with yargs@17 ...

Unfortunately, I think we might want to hold off on dropping Node 10 until I can drop Node 10 on the libraries I manage at Google.

We can implement a less elegant version here:

https://github.com/bcoe/c8/blob/main/lib/commands/report.js#L29

Which I believe should work in the appropriate contexts (we'll want to confirm it works for the `check-coverage` command).

Let's add a `// TODO: comment` to switch to the `middleware` approach as soon as we can drop Node 10?

bcoe:
@gr2m thank you for the contribution \o/

PR #340, opened 2021-12-24
Adding `--extension` option, allowing override of defaults.

brev:
Adding `--extension` option, allowing override of defaults. 

I'm using experimental ESM loaders to have Node.js process `.svelte` files, and now I can run code coverage on them too.

Modified `--all` option to check file extensions against `extension` list
instead of a hard-coded list.

Updated links for default lists in README for `--exclude` and `--extension`.

Tests are passing.

I did not add new tests, as that would require either changing, or adding, a
test runner to use ESM loaders. As the changes here are simple, I thought it
might be overkill. But let me know what you think.

thanks!

##### Checklist


brev:
Oh wow, this is a duplicate of https://github.com/bcoe/c8/pull/331 haha.

bcoe:
@brev thank you for taking the time to make a contribution 👏 

I'm going ahead and landing #331, since it was most of the way there, and just needed to have snapshots updated.

Please don't hesitate to contribute in the future.

PR #343, opened 2021-12-30
fix(config): support configuration inheritance

### LINKED ISSUE #320: Configuration doesn't inherit with `extends` keyword
* **Version**: 7.8.0
* **Platform**: linux mint

This change has broken configuration inheritance:
https://github.com/bcoe/c8/commit/99436ef131c2ab966174b5012fe22e499fb44ccd#diff-40d93e4e5f4b94fc8a34a87a3328651c24e855051bfcdd6944dee76f9c82b8d4L116

> see: http://yargs.js.org/docs/#api-reference-configobject-extends-keyword

custom `configParser` doesn't provide inheritance
https://github.com/bcoe/c8/commit/99436ef131c2ab966174b5012fe22e499fb44ccd#diff-40d93e4e5f4b94fc8a34a87a3328651c24e855051bfcdd6944dee76f9c82b8d4R15

bcoe:
Use the `applyExtends` helper that yargs exposes to support extending config.

Fixes #320


test/integration.js_10.snap
```diff
@@ -337,12 +322,12 @@ exports[`c8 report generates report from existing temporary files 1`] = `
 ",--------------------------|---------|----------|---------|---------|--------------------------------
 File                      | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s              
 --------------------------|---------|----------|---------|---------|--------------------------------
-All files                 |    73.6 |    59.03 |    62.5 |    73.6 |                                
+All files                 |   73.72 |    59.03 |    62.5 |   73.72 |                                
  bin                      |   78.84 |       60 |   66.66 |   78.84 |                                
   c8.js                   |   78.84 |       60 |   66.66 |   78.84 | 22,27-29,32-33,41-43,50-51     
- lib                      |   78.19 |    54.38 |      72 |   78.19 |                                
+ lib                      |   78.32 |    54.38 |      72 |   78.32 |
```
bcoe, line 328: It's weird to me that this coverage includes information from `c8` itself in Node 10, but that Node 12+ does not.

I checked the raw output from Node.js, and it's an actual discrepancy between the two outputs. We'll be dropping Node 10 soon, so perhaps this is okay.

Issue #325, opened 2021-09-09
Incorrect branch coverage when loaders used


coderaiser:
`c8` shows uncovered lines, but the whole file is covered.

![image](https://user-images.githubusercontent.com/1573141/132669108-53525a81-6bb5-4a72-aa07-81360ce1b5c5.png)

Version: output of `node -v` 16.9.0
Platform: output of `uname -a` Darwin

* **Version**: latest
* **Platform**: mac os

Repository https://github.com/coderaiser/c8-reproduce

When I'm using [loaders](https://nodejs.org/api/esm.html#esm_transformsource_source_context_defaulttransformsource) to mock `imports` the coverage I see is broken.

Code:

```js
import {
    readFile,
} from 'fs/promises';

import {
    execSync,
} from 'child_process';

export default (a, b, c) => {
    if (a)
        return readFile();

      if (c)
          return execSync();

      return 'd';
};
```

Tests:

```js
import {createMockImport} from 'mock-import';
import {
    test,
    stub,
} from 'supertape';

const {mockImport, reImport, stopAll} = createMockImport(import.meta.url);

test('changelog: a', async (t) => {
    mockImport('fs/promises', {
        readFile: stub().returns('a'),
    });
    const fn = await reImport('./changelog.js');
    stopAll();

    t.equal(fn.default(1), 'a');
});

test('changelog: c', async (t) => {
    mockImport('child_process', {
        execSync: stub().returns('c'),
    });

    const fn = await reImport('./changelog.js');
    stopAll();

    t.equal(fn.default(0, 0, 1), 'c');
});

test('changelog: d', async (t) => {
    const fn = await import('./changelog.js?count=4');

    t.equal(fn.default(0, 0, 0, 1), 'd');
});
```

What [mock-import](https://github.com/coderaiser/mock-import) does is converts source to:

```js
const {
    readFile: readFile
} = global.__mockImportCache.get('fs/promises');

import {
    execSync,
} from 'child_process';

export default (a, b, c) => {
    if (a)
        return readFile();
     
      if (c)
          return execSync();
      
      return 'd';
};
```
And `imports` it as `./changelog.js?count=1` on first test, then on second test:

```js
import {
    readFile
} from 'fs/promises'

const {
    execSync,
} = global.__mockImportCache.get('fs/promises');

export default (a, b, c) => {
    if (a)
        return readFile();
     
      if (c)
          return execSync();
      
      return 'd';
};
```

[...]

PR #345, opened 2022-01-02
feature: report: ability to use dynamic source maps for `?count=x` imported files

coderaiser:
This is example of changes that can be made to get support of loaded and transformed files (#325).

As I understand, the problem can be somewhere in [v8-to-instanbul](https://github.com/istanbuljs/v8-to-istanbul/blob/v8.1.0/lib/v8-to-istanbul.js#L108-L185) when we are going to apply coverage. The thing is files sizes is different, and offsets different too, and they should be converted somehow 🤔. And there is conversion but it returns `Infinity` and `NaN` for some reason. 

##### Checklist

### DIFF
--- a/lib/report.js
+++ b/lib/report.js
@@ -93,7 +93,7 @@ class Report {
     for (const v8ScriptCov of v8ProcessCov.result) {
       try {
         const sources = this._getSourceMap(v8ScriptCov)
-        const path = resolve(this.resolve, v8ScriptCov.url)
+        const path = resolve(this.resolve, fileURLToPath(v8ScriptCov.url))
         const converter = v8toIstanbul(path, this.wrapperLength, sources, (path) => {
           if (this.excludeAfterRemap) {
             return !this.exclude.shouldInstrument(path)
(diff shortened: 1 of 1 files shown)


bcoe:
@coderaiser I would love to close out some of these long standing PRs. Any chance I could bother you to write a test for this, and we can get it merged?

coderaiser:
Ia have no idea how to test this, if you know the best way for this, will be amazing to have support of dynamic imports merged with this or any other PR.

bcoe:
@coderaiser mind rebasing and we can go from there? 

I think what we could potentially do would be to try to write a test based on a fixture of output from Node.js

coderaiser:
Just rebased

bcoe:
@coderaiser looks like something is weird here, quite a few failing tests.

coderaiser:
The thing is this is not a full solution, I didn’t figure out the full way of fixing this behaviour, you can take a look at repository it   https://github.com/coderaiser/c8-reproduce, the problem in resulting coverage file that should be merged for files `a.js`and `a.js?any-quary`, so file can be re-imported in tests and it was mirrored in coverage files.

PR #349, opened 2022-01-03
build: Add Node 17 to matrix

make-github-pseudonymous-again:
Tests are passing on my machine:

**Version**: `v17.3.0`
**Platform**: `Linux 5.15.12-arch1-1 #1 SMP PREEMPT Wed, 29 Dec 2021 12:04:56 +0000 x86_64 GNU/Linux`

##### Checklist

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [10, 12, 14, 16]
+        node: [10, 12, 14, 16, 17]
     steps:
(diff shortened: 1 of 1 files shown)


bcoe:
actually, maybe let's add 17 given that we want to catch coverage regressions early for `c8` -- what do you think? For other libraries I would only add even numbers

bcoe (approved):

make-github-pseudonymous-again:
Yep. I am mainly interested in this because 17 is what runs on Arch. Does it make sense to cover all LTS releases and latest? I think you are right. But maybe it depends, as always. Given the scope of `c8` I think it is justified.

PR #352, opened 2022-01-19
feat: exclude unexpected file in sourcemap result

bvanjoi:
##### Checklist

##### Background

Sometimes we need to do coverage on the bundled code file, which contains a lot of useless code, for example `node_modules`.

##### Feature 

This PR ensure useless files are remove by `this.exclude`.


bvanjoi:
Such as test case in my PR, which try run `c8 bundle.js`, and its output only contains `index.js`.

If we do not `exclude`, the result will contains `index.js` and `node_modules/path-exists`:

image

bcoe:
@bvanjoi if I'm understanding correctly, this specifically addresses the issue of `1:many` source maps, where a single bundled file might expand out into a few files, some of which are excluded?

bvanjoi:
> @bvanjoi if I'm understanding correctly, this specifically addresses the issue of `1:many` source maps, where a single bundled file might expand out into a few files, some of which are excluded?

Yeah, It is. Maybe I need to update snapshot.

bvanjoi:
image

Differents node version had differents `branch` coverage.

- `<= 12` tag branch coverage is 100.
- but `>= 14` tag it is 0.

This is probably due to the coverage file generated by node(left is node 14, and right is node12)

image

bcoe:
@bvanjoi the culprit is the ` ` between the closing `}` and the `catch` on line `55`.

```
> code.slice(1936, 1949)
' catch (err) '
```

@glromeo I don't suppose any of the work you've been doing in `v8-to-istanbul` happens to solve this edge-case? (I don't expect it does, since the problem is mid-line).

glromeo:
@bcoe I checked out @bvanjoi branch and I verified that my https://github.com/istanbuljs/v8-to-istanbul/pull/179 fixes that failing test.
Also https://github.com/istanbuljs/v8-to-istanbul/pull/180 fixes that error but for a different reason (in general it goes one character short and picks the previous sourcemapped point if one is absent)

bcoe:
@bvanjoi could I bother you to update to the latest version of `v8-to-istanbul` on your branch, to see if it addresses some of your failing tests.

bvanjoi:
> @bvanjoi could I bother you to update to the latest version of `v8-to-istanbul` on your branch, to see if it addresses some of your failing tests.

I had upgrade `v8-to-istanbul` to `8.1.1`, but it still failed in `node 12`, succeed in `node14`, `node16`:

image

bvanjoi:
> @bvanjoi give things a shot with the latest fixes to `v8-to-istanbul`?

It still had some problems in `node 10` and `node 12` 😢

image

image

bcoe:
@bvanjoi, it seems like @glromeo's upstream tweaks might be necessary for this PR to work on both Node 12 and Node 10.

PR #361, opened 2022-02-18
chore: enforce LF

XhmikosR:
On Windows `autocrlf` is `true` by default so this enforces LF regardless of the git settings.

##### Checklist

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Enforce Unix newlines
+* text=auto eol=lf


bcoe (commented):
Won't this make it harder for us to catch broken Windows behavior, as our tests on Windows won't notice that we haven't taken into account `\r\n`?

XhmikosR:
It doesn't affect anything as things are right now (as you can see by the tests). In theory, it could, but this has only to do with the source file line endings which in fact keep marking the test snapshot as modified on Windows due to this.

PR #368, opened 2022-02-18
chore: update dependencies

XhmikosR:
##### Checklist

TODO:

- [ ] figure out the Node.js 10 failure (on my dev branch this is passing: https://github.com/XhmikosR/c8/runs/5245909188?check_suite_focus=true so maybe a test needs to be adapted?)

Closes #356 along the way.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,59 +10,68 @@
       "license": "ISC",
       "dependencies": {
         "@bcoe/v8-coverage": "^0.2.3",
-        "@istanbuljs/schema": "^0.1.2",
+        "@istanbuljs/schema": "^0.1.3",
         "find-up": "^5.0.0",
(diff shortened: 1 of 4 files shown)


bcoe:
It seems odd to me that this would change snapshot output significantly, do you know which dependency update is doing this?

XhmikosR:
Unfortunately, I don't know which dependency exactly is causing this.

XhmikosR:
Could you try updating the Node.js 10 snapshot? It might be just it's behind.

bcoe:
see: https://github.com/bcoe/c8/issues/385

I've created a separate PR that updates deps.

XhmikosR:
So, this leaves `v8-to-istanbul` which could have caused the issue.

PR #370, opened 2022-02-18
chore: assorted lint tweaks

XhmikosR:
* switch to strict mode
* use `for...of`
* use `Array.includes` over `indexOf`
* use `startsWith` over regex
* return early in a few cases

Notes:

* non-whitespace diff: https://github.com/bcoe/c8/pull/370/files?w=1
* some of the changes are subjective; happy to adapt them
* the last patch seems cleaner to me, but happy to revert
* I haven't done any benchmarks, if someone can guide me, happy to include them

##### Checklist

### DIFF
--- a/test/integration.js.snap
+++ b/test/integration.js.snap
@@ -157,43 +157,6 @@ All files   |   83.33 |      100 |   66.66 |   83.33 |
 ,"
 `;
 
-exports[`c8 can allow for files outside of cwd 1`] = `
-",hi
--------------------|---------|----------|---------|---------|-------------------
-File               | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
--------------------|---------|----------|---------|---------|-------------------
-All files          |     100 |      100 |     100 |     100 |                   
- multidir1         |     100 |      100 |     100 |     100 |                   
-  file1.js         |     100 |      100 |     100 |     100 |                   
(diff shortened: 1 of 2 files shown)


lib/report.js
```diff
@@ -1,15 +1,19 @@
...
 const Exclude = require('test-exclude')
 const libCoverage = require('istanbul-lib-coverage')
 const libReport = require('istanbul-lib-report')
 const reports = require('istanbul-reports')
-const { readdirSync, readFileSync, statSync } = require('fs')
-const { isAbsolute, resolve, extname } = require('path')
-const { pathToFileURL, fileURLToPath } = require('url')
-const getSourceMapFromFile = require('./source-map-from-file')
 // TODO: switch back to @c88/v8-coverage once patch is landed.
+const { mergeProcessCovs } = require('@bcoe/v8-coverage')
```
XhmikosR, line 12: Happy to move any `require`s back if you think they affect load time a lot. From my local load benchmark, it seems there's a small penalty ~3-5ms average.

bcoe:
These changes touch quite a bit of code and are purely stylistic.

The goal of using `standard` is that we don't make PRs like this, rather we opt for the stylistic rules currently enforced by standard.

Having said this, would happily accept a patch updating `standard` if we've fallen behind.

XhmikosR:
These are not only stylistic changes and they weren't made automatically.

No worries, though.

PR #372, opened 2022-02-18
refactor!: remove pre-Node.js 10.x guards from code

XhmikosR:
Node.js < 10 is not supported by the current main AFAICT, so these are probably leftovers and can be removed.

##### Checklist

- [ ] rebase if it conflicts with #370


bcoe (changes requested):
I'm open to doing this, but I think we should do so for the next major change, when we drop Node 10; as this has the potential to be breaking for some people -- if anyone is still using c8 on older versions of Node.js.

bcoe:
closing for now, let's revisit when we drop support for Node 10.

PR #388, opened 2022-04-06
fix(perf): cache `this.exclude.shouldInstrument` for improved performance

sapphi-red:
This PR drastically improves performance for large codebases.
Below is the result of total time collecting coverage for [my project](https://github.com/traPtitech/traQ_S-UI) which uses c8 internally in vitest.

|    |Before|After|Diff|
|---|---|---|---|
|with `all=false`|37.056s|7.126s|-29.93s (-80%)|
|with `all=true`|33.515s|2.849s|-30.66s (-91%)|

This result was measured by adding `console.time` and `console.timeEnd` around here.
https://github.com/bcoe/c8/blob/ae1f5e599476ae38b09b92c9f7db837e79cd1693/lib/report.js#L66-L78

This is a bit off topic but FYI [this `this.exclude.globSync(workingDir)` call](https://github.com/bcoe/c8/blob/ae1f5e599476ae38b09b92c9f7db837e79cd1693/lib/report.js#L193) takes 3.6s totally.
So there is a room to improve more if I can change implementation inside `test-exclude`.

refs: https://github.com/vitest-dev/vitest/issues/1075#issuecomment-1090139760

##### Checklist

### DIFF
--- a/lib/report.js
+++ b/lib/report.js
@@ -44,6 +44,7 @@ class Report {
       excludeNodeModules: excludeNodeModules
     })
     this.excludeAfterRemap = excludeAfterRemap
+    this.shouldInstrumentCache = new Map()
     this.omitRelative = omitRelative
     this.sourceMapCache = {}
     this.wrapperLength = wrapperLength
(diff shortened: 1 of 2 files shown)


bcoe (approved):
This seems like a great improvement, thank you.

sapphi-red:
@bcoe Thank you for reviewing! I have updated snapshot for Node 10.

bcoe (approved):

PR #405, opened 2022-05-27
Add support for projectRoot option

rasimx:
When code is covered with tests in a mono repository, where the frontend and backend store lcov.info files separately, we need the ability to specify the relative path to the repository root.  The istanbul package provides this feature through the [projectRoot](https://github.com/istanbuljs/istanbuljs/pull/492) option

##### Checklist


bcoe:
@rasimx mind running `test:snap` after `nvm use 10` as well?

PR #413, opened 2022-07-17
docs: add missing description for --exclude-after-remap

ArrayZoneYour:
Reference:

https://github.com/istanbuljs/nyc#source-map-support-for-pre-instrumented-codebases

https://github.com/bcoe/c8/pull/293

##### Checklist


README.md
```diff
@@ -53,6 +54,17 @@ By supplying `--all` to c8, all files in directories specified with `--src` (def
 and `--exclude` flag checks, will be loaded into the report. If any of those files remain uncovered they will be factored
 into the report with a default of 0% coverage.
 
+## Source-Map support for pre-instrumented codebases
+
+If you opt to pre-instrument your source-code (rather than using a just-in-time
+transpiler like [`@babel/register`]) c8 supports both inline source-maps and
+`.map` files.
+
+_Important: If you are using c8 with a project that pre-instruments its code,
```
bcoe, line 63: It would be useful to give an example of when we would want to set `--exclude-after-remap` to `true`, vs, when we would want to leave it `false`, rather than this blurb.

README.md
```diff
@@ -53,6 +54,17 @@ By supplying `--all` to c8, all files in directories specified with `--src` (def
 and `--exclude` flag checks, will be loaded into the report. If any of those files remain uncovered they will be factored
 into the report with a default of 0% coverage.
 
+## Source-Map support for pre-instrumented codebases
```
bcoe, line 57: Perhaps instead:

## Pre-instrumented vs., just-in-time instrumented codebases

### Source map files, vs., inline source maps

Just-in-time instrumented codebases will often insert source maps inline with the `.js` code they generate at runtime (_as an example, @babel-register/register can be configured to insert a source map foote).

Pre-instrumented codebases, e.g., running `tsc` to generate  `.js` in a build folder, may generate either inline source maps, or a separate `.map` file stored on disk.

`c8` can handle loading both types of source maps.

### Exclude after remap

... explain here why you when you would set the setting to `true`, vs., `false`.

bcoe (changes requested):
Thanks for the contribution 👏 

I left some suggestions about how we might better explain the difference between inline vs., on disk source maps, and the setting `--exclude-after-remap`.

ArrayZoneYour:
@bcoe Thanks for suggestions and review 😀. Update the explanations and be willing to make some other changes if it is required.

README.md
```diff
@@ -53,6 +54,22 @@ By supplying `--all` to c8, all files in directories specified with `--src` (def
 and `--exclude` flag checks, will be loaded into the report. If any of those files remain uncovered they will be factored
 into the report with a default of 0% coverage.
 
+## Pre-instrumented versus just-in-time instrumented codebases
+
+### Source map files versus inline source maps
+
+Just-in-time instrumented codebases will often insert source maps inline with the `.js` code they generate at runtime (e.g, `@babel/register` can be configured to insert a source map foote).
```
bcoe, line 61:
```suggestion
Just-in-time instrumented codebases will often insert source maps inline with the `.js` code they generate at runtime (e.g, `@babel/register` can be configured to insert a source map footer).
```

README.md
```diff
@@ -53,6 +54,22 @@ By supplying `--all` to c8, all files in directories specified with `--src` (def
...
+
+Pre-instrumented codebases, e.g., running `tsc` to generate `.js` in a build folder, may generate either inline source maps, or a separate `.map` file stored on disk.
+
+`c8` can handle loading both types of source maps.
+
+### Exclude after remap
+
+For independent projects, feel free to set `--exclude-after-remap` as `true` or `false`, the only difference is that the exclusions are applied to _**source**_ scripts or _**executable**_ scripts.
+
+For huge 'shared codebase' such as monorepo, your test suites can use common libs as workspace node_modules and _**executable**_ scripts is a single bundle, set `--exclude-after-remap` to true may help to set exclusions more precisely.
```
bcoe, line 71:
```suggestion
Depending on the size and configuration of your project, it may be preferable to apply exclusion logic either before or after source-maps are used to remap compiled to original source files.

`--exclude-after-remap` is used to control this behaviour.
```

README.md
```diff
@@ -53,6 +54,22 @@ By supplying `--all` to c8, all files in directories specified with `--src` (def
 and `--exclude` flag checks, will be loaded into the report. If any of those files remain uncovered they will be factored
 into the report with a default of 0% coverage.
 
+## Pre-instrumented versus just-in-time instrumented codebases
```
bcoe, line 57:
```suggestion
## SourceMap Support

`c8` can handle source-maps, for remapping coverage from generated code to original source files (_useful for TypeScript, JSX, etc).
```

bcoe (changes requested):
Left a couple more editing suggestions, does this get at what you were picturing?

README.md
ArrayZoneYour: LGTM

README.md
```diff
@@ -53,6 +54,24 @@ By supplying `--all` to c8, all files in directories specified with `--src` (def
 and `--exclude` flag checks, will be loaded into the report. If any of those files remain uncovered they will be factored
 into the report with a default of 0% coverage.
 
+## SourceMap Support
+
+`c8` can handle source-maps, for remapping coverage from generated code to original source files (_useful for TypeScript, JSX, etc).
```
bcoe, line 59:
```suggestion
`c8` can handle source-maps, for remapping coverage from generated code to original source files (_useful for TypeScript, JSX, etc_).
```

bcoe (approved):

bcoe:
@ArrayZoneYour thank you for the contribution.

PR #415, opened 2022-08-19
Add --projectRoot option

rasimx:
##### Checklist


AriPerkkio (commented):
Should this be passed to [`test-exclude`](https://github.com/istanbuljs/test-exclude#optionscwd) as `cwd` as well?

https://github.com/bcoe/c8/blob/7f1069dcce6821a6794557b39d6a13f620c64bad/lib/report.js#L39-L45

rasimx:
> Should this be passed to [`test-exclude`](https://github.com/istanbuljs/test-exclude#optionscwd) as `cwd` as well?
> 
> https://github.com/bcoe/c8/blob/7f1069dcce6821a6794557b39d6a13f620c64bad/lib/report.js#L39-L45
I do not think so. This option affects only the paths in the lcov file.

lib/parse-args.js
```diff
@@ -152,6 +152,11 @@ function buildYargs (withCommands = false) {
       describe: 'supplying --allowExternal will cause c8 to allow files from outside of your cwd. This applies both to ' +
         'files discovered in coverage temp files and also src files discovered if using the --all flag.'
     })
+    .options('project-root', {
+      default: undefined,
+      type: 'string',
+      describe: 'specifies which points point to the root of the mono repository'
```
AriPerkkio, line 158: I wonder if this option overlaps with the undocumented `resolve` option. Have you tried using it as the `projectRoot` in istanbul-reporter options?

```js
projectRoot: this.resolve || process.cwd(),
```

With the current changes it feels like you'd need to keep `projectRoot` and `resolve` in sync manually.

https://github.com/bcoe/c8/blob/7f1069dcce6821a6794557b39d6a13f620c64bad/lib/parse-args.js#L136-L139

https://github.com/bcoe/c8/blob/7f1069dcce6821a6794557b39d6a13f620c64bad/lib/report.js#L97

rasimx: To be honest, I don't understand why the 'resolve' option is needed. Whatever I specify there, it does not affect the final files.

rasimx: I tried using this, but it didn't help me.

rasimx: projectRoot only affects paths in lcov files. We are currently using a fork in a working project and it works fine.

AriPerkkio: I did some testing with the `this.resolve` option now, and it indeed seems to have no effect at all. The `v8ScriptCov.url` paths are absolute, not relative. No idea what that API is supposed to do.

AriPerkkio:
> I do not think so. This option affects only the paths in the lcov file.

I still think this same option should be used in `test-exclude` as `cwd`. This is what `nyc` does, and this project seems to mimic its API quite well. Let's wait for project maintainers to make the final call though. 

`nyc` has `cwd` option, which is used as `projectRoot` for `istanbul-lib-report` and as `cwd` for `test-exclude`:

> [You can change the project root directory with the `--cwd` option.](https://github.com/istanbuljs/nyc/#setting-the-project-root-directory)

- https://github.com/istanbuljs/nyc/blob/v15.1.0/index.js#L69
- https://github.com/istanbuljs/nyc/blob/v15.1.0/index.js#L465

This feature is also something that is needed in `vitest` in order to change project root: https://github.com/vitest-dev/vitest/issues/1902.

bcoe:
Hello @rasimx could you provide a description in your PR of the problem you're trying to solve?

I believe the `resolve` functionality rewrites the output from v8 to resolve under an alternate path, allowing you to remove leading directories, etc. Whereas passing `projectRoot` to the Reporter is only used by a couple reporters (clover and cobertura, see [search](https://github.com/gotwarlost/istanbul/search?q=projectRoot)).

lib/report.js
```diff
@@ -74,7 +76,8 @@ class Report {
       reports.create(_reporter, {
         skipEmpty: false,
         skipFull: this.skipFull,
-        maxCols: process.stdout.columns || 100
+        maxCols: process.stdout.columns || 100,
+        projectRoot: this.projectRoot
```
AriPerkkio, line 80: @rasimx looks like an alternative way to pass this option was implemented in https://github.com/bcoe/c8/pull/423.

It's included in [`c8@7.13.0`](https://github.com/bcoe/c8/pull/428).

bcoe:
> @rasimx looks like an alternative way to pass this option was implemented in https://github.com/bcoe/c8/pull/423.

I would prefer that we expose this reporterOptions configuration so that it can be configured in package configuration, or with `--reporter-options.projectRoot` -- reasoning being that the options available for the reporter sometimes have names that are a bit confusing because they sound like top level configuration for `c8` itself (this is why I was initially confused by this PR).

 @rasimx closing this for now, since I did't hear back from you with my last update. But, I'd be happy to land this work in the future if you're willing to update based on this suggestion.

PR #423, opened 2022-10-21
feat: support passing reporter options

elliottsj:
For example:

```
{
  reporter: ['text'],
  reporterOptions: {
    text: { file: 'coverage.txt' },
  }
}
```

This enables overriding the `file` option which normally defaults to `null`, so text reporters can write to a file rather than stdout: https://github.com/istanbuljs/istanbuljs/blob/fb8cb4bb99c9e5d36c8b011338d2ab5d3c72ff9c/packages/istanbul-reports/lib/text/index.js#L233-L237

##### Checklist

### DIFF
--- a/lib/report.js
+++ b/lib/report.js
@@ -19,6 +19,7 @@ class Report {
     excludeAfterRemap,
     include,
     reporter,
+    reporterOptions,
     reportsDirectory,
     tempDirectory,
     watermarks,
@@ -32,6 +33,7 @@ class Report {
     excludeNodeModules
   }) {
     this.reporter = reporter
+    this.reporterOptions = reporterOptions || {}
     this.reportsDirectory = reportsDirectory
     this.tempDirectory = tempDirectory
     this.watermarks = watermarks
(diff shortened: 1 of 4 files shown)


bcoe:
This seems like a very reasonable addition to the API 😄

bcoe:
@elliottsj could I bother you to run the Node 10 snapshots, you can do so by switching to Node 10 using `nvm`.

bcoe (approved):

bcoe:
@elliottsj thank you for the contribution 👌 , a good follow up would be docs if you're so inclined.

elliottsj:
Ah sorry, thank you for updating the snapshots!

I'll look at writing docs

rosingrind:
@bcoe @elliottsj looks like these changes doesn't work on 7.13.0: `reporterOptions` is constantly a blank object. Isn't  `buildYargs()` missing an entry for `reporterOptions` here?

https://github.com/bcoe/c8/blob/2f36fe9d00bf2f3b869ce890ed11416304c903fe/lib/parse-args.js#L23-L28

koshic:
+1, `reporterOptions` parameter is empty.

Issue #431, opened 2022-11-21
Add support for yaml/.js configuration files


bcoe:
We've had a few requests for yaml and .js configuration files:

https://github.com/bcoe/c8
https://github.com/bcoe/c8/issues/420

---

One issue with .yaml is that the available yaml parsers are massive, 10x the size of c8 itself.

.js configuration would perhaps be less unwieldy. We should look at how mocha approaches this (as I think there are potentially security pitfalls, we shouldn't just go with the naive approach).

kf6kjg:
Parsers may be massive, but they are also already deeply embedded in the ecosystem. E.g. a lot of the dev packages a user will install already include such. 

That said you could make such a parser an optional dependency. Then the user install it if the user wants to configure that way.

As to using javascript, not my preference unless the system can also accept Typescript. That has been a difficult proposition in most projects I've interacted with.

mcknasty:
Hi @bcoe and @kf6kjg  

I just put in a #436 for this feature request. Need maintainer to review before starting the workflow.

Thanks!

wellwelwel:
@bcoe, I'd love help with this 🤝

> Recently I got an issue named `“Coverage 100%” badge is misleading`. Being able to comment in `.jsonc` or a `.js` would be great.

---

### Parsing JSON and JSONC at the same time

I created a smaller package (**jsonc.min**) that is a single-file project to ensure a safe approach in parse **JSON** and **JSONC** files independent of the file extension _(e.g., `.c8rc`, `.c8rc.json`, `.c8rc.jsonc`)_ with ~9kB and zero `regex` or external dependencies:

<a href="https://pkg-size.dev/jsonc.min"><img src="https://pkg-size.dev/badge/install/9220" title="Install size for jsonc.min"></a>

> I'd also have no problem including the source code from **jsonc.min** directly without adding the dependency to **c8**.

---

I'd like to share a real example to dynamically accept both `.json`, `.jsonc`, `.js` and `.cjs`, based on a project I maintain:

```js
import { cwd } from 'node:process';
import { normalize, join } from 'node:path';
import { readFile } from 'node:fs/promises';
import { JSONC } from 'jsonc.min';

const processCWD = cwd();

export const getConfigs = async (
  customPath?: string
) => {
  const expectedFiles = customPath
    ? [customPath]
    : [
        '.c8rc',
        '.c8rc.json',
        '.c8rc.jsonc',
        '.c8rc.js',
[...]

PR #436, opened 2022-12-04
Adding additional configuration file options

mcknasty:
# **[Pull Request 436](https://github.com/bcoe/c8/pull/436)**

## Commit Message

commit 22a9c08a01a219a97e9768be7b211a6793c22956
Author: The Nasty <tmckeown@gmail.com>
Date:   Sat Jan 20 13:51:03 2024 -0500

    Additional Configuration File Support (#436)

    feat: Adding additional configuration file options
    test: unit test for config file detection
    test: unit test for config files pass via the cli
    test: unit test case for early exit of hideInstrumenteeArgs
    test: unit test for Node.js argument handling
    fix: bug in spawning of c8 process that dropped coverage significantly

## **Following the Contributions Recommendations [here.](https://github.com/bcoe/c8/blob/main/CONTRIBUTING.md)**

1. [x] Make sure you have installed the latest version of Node.js
2. [x] Fork this project on GitHub and clone your fork locally
3. [x] Create local branches to work within. These should also be created directly from the main branch.  Local Fork [here.](https://github.com/mcknasty/c8)
4. [x] Make your changes.
5. [x] Run tests to make sure all is okay (everything should pass except the snapshot).
   1. A [complete log](https://gist.github.com/mcknasty/6315f83810db7e88036431534c312859#file-initial-test-txt) of initial test results.
   2. As instructed, ignore snapshot failures. [17 pending](https://gist.github.com/mcknasty/6315f83810db7e88036431534c312859#file-initial-test-txt-L193)
   3. [120 passing](https://gist.github.com/mcknasty/6315f83810db7e88036431534c312859#file-initial-test-txt-L192) in [53 seconds](https://gist.github.com/mcknasty/6315f83810db7e88036431534c312859#file-initial-test-txt-L192)
6. [x] Now update the snapshot.
   1. A [complete log](https://gist.github.com/mcknasty/6315f83810db7e88036431534c312859#file-snapshot-test-txt) of snapshot test results.
   2. [120 passing](https://gist.github.com/mcknasty/6315f83810db7e88036431534c312859#file-snapshot-test-txt-L196) in [52 seconds](https://gist.github.com/mcknasty/6315f83810db7e88036431534c312859#file-snapshot-test-txt-L196)
7. [x] If all is passing, commit your changes. The log of the commit can be found [here.](https://github.com/bcoe/c8/pull/436/commits/22a9c08a01a219a97e9768be7b211a6793c22956)
8. [x] As a best practice, once you have committed your changes, it is a good idea to use git rebase (not git merge) to synchronize your work with the main repository.
9. [x] Run tests again to make sure all is okay.
    1. A [complete log](https://gist.github.com/mcknasty/6315f83810db7e88036431534c312859#file-final-test-txt) of the final test results.
    2. [120 passing](https://gist.github.com/mcknasty/6315f83810db7e88036431534c312859#file-final-test-txt-L192) in [54 seconds](https://gist.github.com/mcknasty/6315f83810db7e88036431534c312859#file-final-test-txt-L192)
10. [x] Push
11. [x] Open the pull request, see details in the template.
12. [ ] Make any necessary changes after review.

## **Unit Tests Results**

|   Test Type     |  PASS          |  Tests Passed   |  Tests Pending  |    Time      |
|:---------------:|:--------------:|:---------------:|:--------------:|:------------:|
| [Initial Test](https://gist.github.com/mcknasty/6315f83810db7e88036431534c312859#file-initial-test-txt)       | :heavy_check_mark: | [120 passing](https://gist.github.com/mcknasty/6315f83810db7e88036431534c312859#file-initial-test-txt-L192)        | [17 pending](https://gist.github.com/mcknasty/6315f83810db7e88036431534c312859#file-initial-test-txt-L193)       | [53 seconds](https://gist.github.com/mcknasty/6315f83810db7e88036431534c312859#file-initial-test-txt-L192)       |
| [Snapshot Test](https://gist.github.com/mcknasty/6315f83810db7e88036431534c312859#file-snapshot-test-txt)       | :heavy_check_mark: | [120 passing](https://gist.github.com/mcknasty/6315f83810db7e88036431534c312859#file-snapshot-test-txt-L196)        | [17 pending](https://gist.github.com/mcknasty/6315f83810db7e88036431534c312859#file-snapshot-test-txt-L197)       | [52 seconds](https://gist.github.com/mcknasty/6315f83810db7e88036431534c312859#file-snapshot-test-txt-L196)       |
| [Final Test](https://gist.github.com/mcknasty/6315f83810db7e88036431534c312859#file-final-test-txt)       | :heavy_check_mark: | [120 passing](https://gist.github.com/mcknasty/6315f83810db7e88036431534c312859#file-final-test-txt-L192)        | [17 pending](https://gist.github.com/mcknasty/6315f83810db7e88036431534c312859#file-final-test-txt-L193)       | [54 seconds](https://gist.github.com/mcknasty/6315f83810db7e88036431534c312859#file-final-test-txt-L192)       |

* Test completed in Node version v18.19.0

## **Node Version Testing Matrix**

|   Node Version  |  PASS          |  Tests Passed   |  Tests Pending  |    Time      |
|:---------------:|:--------------:|:---------------:|:--------------:|:------------:|
| [14.21.3](https://gist.github.com/mcknasty/104af724ffd8dff4c4a5f60e54eed1da#file-v14-21-3-test-txt)       | :heavy_check_mark: | [120 passing](https://gist.github.com/mcknasty/104af724ffd8dff4c4a5f60e54eed1da#file-v14-21-3-test-txt-L215)        | [17 pending](https://gist.github.com/mcknasty/104af724ffd8dff4c4a5f60e54eed1da#file-v14-21-3-test-txt-L216)       | [1 minutes](https://gist.github.com/mcknasty/104af724ffd8dff4c4a5f60e54eed1da#file-v14-21-3-test-txt-L215)       |
| [16.20.2](https://gist.github.com/mcknasty/104af724ffd8dff4c4a5f60e54eed1da#file-v16-20-2-test-txt)       | :heavy_check_mark: | [120 passing](https://gist.github.com/mcknasty/104af724ffd8dff4c4a5f60e54eed1da#file-v16-20-2-test-txt-L216)        | [17 pending](https://gist.github.com/mcknasty/104af724ffd8dff4c4a5f60e54eed1da#file-v16-20-2-test-txt-L217)       | [57 seconds](https://gist.github.com/mcknasty/104af724ffd8dff4c4a5f60e54eed1da#file-v16-20-2-test-txt-L216)       |
| [18.19.0](https://gist.github.com/mcknasty/104af724ffd8dff4c4a5f60e54eed1da#file-v18-19-0-test-txt)       | :heavy_check_mark: | [120 passing](https://gist.github.com/mcknasty/104af724ffd8dff4c4a5f60e54eed1da#file-v18-19-0-test-txt-L212)        | [17 pending](https://gist.github.com/mcknasty/104af724ffd8dff4c4a5f60e54eed1da#file-v18-19-0-test-txt-L213)       | [1 minutes](https://gist.github.com/mcknasty/104af724ffd8dff4c4a5f60e54eed1da#file-v18-19-0-test-txt-L212)       |
| [20.11.0](https://gist.github.com/mcknasty/104af724ffd8dff4c4a5f60e54eed1da#file-v20-11-0-test-txt)       | :heavy_check_mark: | [120 passing](https://gist.github.com/mcknasty/104af724ffd8dff4c4a5f60e54eed1da#file-v20-11-0-test-txt-L212)        | [17 pending](https://gist.github.com/mcknasty/104af724ffd8dff4c4a5f60e54eed1da#file-v20-11-0-test-txt-L213)       | [1 minutes](https://gist.github.com/mcknasty/104af724ffd8dff4c4a5f60e54eed1da#file-v20-11-0-test-txt-L212)       |

Duplicate of #447
References #448

- edited 2022-12-05 4:58 AM EST. I noticed an error in my last patch, so I updated it.
- edited 2023-02-20 12:41 PM EST. Updated pull request readme to the new format and linked existing issues.
- edited 2023-07-28 11:50 AM EST. Adding additional unit tests for each of the config file options.
- edited 2023-07-29 1:14 PM EST. Adding test case for loading invalid config files and updating readme to clarify ESM support.
- edited 2024-01-08 3:17 PM EST.  Fixing error that was causing code coverage drops.  Changing Commit Message to include omitted items.
- edited 2024-01-20 3:09 PM EST.  Added test name for chia-snapshots.  Skipped tests that require (#517) to be merged.  Updated an error handling test.


mcknasty:
Feature Request #431

kf6kjg (approved):
What I've been wanting to make my switch over from Istanbul cleaner! :)

I know my review doesn't count for much, but here it is anyway. :D

kf6kjg:
And now I trip across this again, this time matching up a project that uses the `.cjs` config style so that it solves a path resolution problem in a monorepo: the CWD is down in a package subfolder, but the paths for the coverage folders etc are at the monorepo root. Fun times. Worked great on ESLint and Prettier, but that because both of them use this config pattern.

On a side note WRT rolling your own like this and using a library like the very well established [cosmiconfig](https://github.com/davidtheclark/cosmiconfig):
- Less code for you to maintain. Just redirect any further config file feature requests or changes to the underlying library.
- You get features by simply upgrading - like the upcoming [ESM config file support](https://github.com/davidtheclark/cosmiconfig/issues/224).
- You do wind up with a fatter dependency tree. However `c8` doesn't exist in a vacuum: In all probability the user will also be installing packages like ESLint and Prettier, both of which use cosmiconfig. In that case the tree doesn't grow in size as the lib is already there.

mcknasty:
While I certainly appreciate all the hard work put into cosmiconfig, the one possible issue I could see is that cosmiconfig recursively walks the directory structure, looking for a pattern-matching configuration file.   While I am uncertain if node_modules is included in this directory walk, it could potentially affect git submodules that also have a C8 dependency.  A simple typo in a config file name could lead to C8 running with an alternative, undesired, configuration.  Additionally, the solution I provided, I suspect is far less code than an additional package, and could potentially slim the dependency tree in certain projects.

Furthermore, I view a code analysis tool like C8, as a defacto standard for anyone serious about producing quality software.  I see it, or other tools like it, as a dependency for many of the popular packages in the node.js ecosystem.  From my perspective, the smaller the better, for popular packages. This would suggest favoring rolling your features rather than additional package dependencies.  

Finally, I was wondering whether @bcoe concern about the [size of YAML parsers](https://github.com/bcoe/c8/issues/431#issue-1458626689) was addressed on this particular patch.

Thank you for your feedback, insight, and testing of this patch @kf6kjg.

stormsilver:
I check on this PR once a week to see if it's been merged. I cannot wait.

lib/parse-args.js
```diff
@@ -188,6 +189,28 @@ function buildYargs (withCommands = false) {
   return yargs
 }
 
+function loadConfigFile (path, ext) {
+  let config = {}
+  const jsExts = ['.js', '.cjs']
```
bcoe, line 194: What about .mjs? We should likely add a test for this too.

lib/parse-args.js
```diff
@@ -188,6 +189,28 @@ function buildYargs (withCommands = false) {
...
+  const ymlExts = ['.yml', '.yaml']
+  if (jsExts.includes(ext)) {
+    config = require(path)
+  } else if (ymlExts.includes(ext)) {
+    config = require('js-yaml').load(readFileSync(path, 'utf8'))
```
bcoe, line 199: Perhaps we could add a `try/catch` here, and print a useful message that explains that `js-yaml` is necessary to handle `yaml` extensions.

bcoe (changes requested):
This is looking like a good start. I think we should add some unit tests for yaml, js, etc.

I'm fine with us adding `js-yaml` as a dev dependency.

lib/parse-args.js
kf6kjg: The same argument could be made for the other cases. That said I think the default error is fine:
```sh
$ node test.js 
node:internal/modules/cjs/loader:1042
  throw err;
  ^

Error: Cannot find module 'js-yaml'
Require stack:
- /tmp/test.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:1039:15)
    at Module._load (node:internal/modules/cjs/loader:885:27)
    at Module.require (node:internal/modules/cjs/loader:1105:19)
    at require (node:internal/modules/cjs/helpers:103:18)
    at Object.<anonymous> (/tmp/test.js:1:1)
    at Module._compile (node:internal/modules/cjs/loader:1218:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1272:10)
    at Module.load (node:internal/modules/cjs/loader:1081:32)
    at Module._load (node:internal/modules/cjs/loader:922:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [ '/tmp/test.js' ]
}

Node.js v18.13.0
```

lib/parse-args.js
kf6kjg: While I'd love to see mjs support, the catch there is that ESM doesn't like getting loaded into a CommonJS process. The other way 'round can be done easily, but I think it's out of scope for this ticket to convert this package to ESM... :P

mcknasty:
> This is looking like a good start. I think we should add some unit tests for yaml, js, etc.
> 
> I'm fine with us adding `js-yaml` as a dev dependency.

@bcoe You can find the unit tests [here in the commit](https://github.com/bcoe/c8/pull/436/commits/ccbad85fec2c0fbc1450ceeb9d7c20d49beda480#diff-aa8ddbaaf379948f2e332b5b4ef3a1c71387520b3a2eca9d5f6aeba377e00bbf)

test/parse-args.js
```diff
@@ -43,6 +43,26 @@ describe('parse-args', () => {
...
+      argv.lines.should.be.equal(98)
+    })
+    it('should resolve be able to resolve config files with .cjs extensions', () => {
+      const argv = buildYargs().parse(['node', 'c8', 'my-app', '--config', require.resolve('./fixtures/config/.nycrc.cjs')])
+      argv.lines.should.be.equal(94)
+    })
+    it('should resolve be able to resolve config files with .json extensions', () => {
+      const argv = buildYargs().parse(['node', 'c8', 'my-app', '--config', require.resolve('./fixtures/config/.nycrc.json')])
+      argv.lines.should.be.equal(96)
+    })
```
mcknasty, line 65: New unit tests for additional configuration files can be [found here](https://github.com/bcoe/c8/pull/436/commits/ccbad85fec2c0fbc1450ceeb9d7c20d49beda480#diff-aa8ddbaaf379948f2e332b5b4ef3a1c71387520b3a2eca9d5f6aeba377e00bbf)

lib/parse-args.js
mcknasty: The js-yml is currently a dependency of other packages in the c8 dependency tree.  So yes I agree that we should add a try/catch, and also explicitly define the dependency in package.json.  Here is the list of packages that c8 depends on that currently rely on js-yaml

``` bin/bash
***************REDACTED********************:~/codez/tmp/c8$ grep -Rnil 'js-yaml' node_modules/ | sed 's@node_modules/@@g' | sed 's@/@ @g' | awk '{print $1}' | uniq
eslint
uri-js
.package-lock.json
.bin
diff
mocha
typescript
@eslint
js-yaml
***************REDACTED********************:~/codez/tmp/c8$
```

lib/parse-args.js
mcknasty: @kf6kjg I rolled a patch last night / this morning to include ESM support.  This unfortunately is a big change IMO, and I have to overcome a race condition.  Are there any examples of setting up yargs with async functions?

The good news is that all unit tests were passing and I got 100% code coverage on it.  Take a look [at the diff here](https://github.com/mcknasty/c8/pull/1/commits/fb65ab76ec6c2bedf50856ddf3a11288cdf34179)

kf6kjg: The advanced docs for yargs shows some: https://github.com/yargs/yargs/blob/main/docs/advanced.md#using-yargs-with-asyncawait

mcknasty: @kf6kjg Ok thank you for the link.  I will continue to work on that other patch.  We can mark this conversation as resolved, that ESM is out of scope for this request.  I'll open a feature request for ESM support once this PR is merged.

lib/parse-args.js
```diff
@@ -155,6 +156,8 @@ function buildYargs (withCommands = false) {
     .pkgConf('c8')
     .demandCommand(1)
     .check((argv) => {
+      // not sure this block is reached
+      /* c8 ignore next 4 */
```
kf6kjg, line 160: While a definite improvement to the coverage score, I think that these comments might not belong in this PR: no adjacent code is being changed, and the addition of the ignore command hides a possible fault. I recommend to remove these two comments, even at the expense of worse code coverage in this PR: better to fix in a dedicated PR where the author and reviewers can go over the details.
(sorry if the above sounds like it was written by a lawyer, just finished watching an episode of JAG...)

lib/parse-args.js
```diff
@@ -188,6 +191,28 @@ function buildYargs (withCommands = false) {
...
+    config = JSON.parse(readFileSync(path, 'utf8'))
+  }
+  return config
+}
+
+function getConfigFileNames () {
+  const files = ['nyc.config.js']
+  const prefixes = ['.c8rc', '.nycrc']
+  const exts = ['.yaml', '.yml', '.js', '.cjs', '.mjs', '.json', '']
+  prefixes.forEach((prefix) => exts.forEach((ext) => files.push(`${prefix}${ext}`)))
```
kf6kjg, line 212: If I'm reading the [find-up docs](https://github.com/sindresorhus/find-up#findupsyncmatcher-options) correctly, the order of this array is the order of precedence if there are multiple config files on disk - the first in the array being the highest priority.
This code therefor creates a priority as follows:
1. `nyc.config.js`
2. `.c8rc.yaml`
3. `.c8rc.yml`
2. `.c8rc.js`
3. `.c8rc.cjs`
3. `.c8rc.mjs`
3. `.c8rc.json`
3. `.c8rc`
&c. into the `.nycrc` varients.

To me this priority list feels odd and looks to be in disagreement with the [precedence set by `nyc`](https://github.com/istanbuljs/load-nyc-config).

package-lock.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "c8",
   "version": "7.13.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 3,
```
kf6kjg, line 4: I'm not sure this was an intentional change, so I'm bringing it to your attention. Lock file version 3 drops backwards support for NPM v6 in favor of those from 7 onwards. I'm in favor of moving forward, but only if the step is intentional and the maintainer agrees with dropping support for NPM v6 - the version that's shipped with NodeJS 14 by default.

kf6kjg (changes requested):
Looking very good so far. I would like to see the tests fleshed out: [code coverage score is by no means a complete picture](https://igorfil.com/posts/2019-10-30-code-coverage-lies/). To be specific I'd like to see tests for:
1. Priority order when multiple valid config files exist in the same folder.
2. Test demonstrating the expected behavior, whatever that is, when the higher priority config file contains an error but the lower one does not.
3. Tests that prove that each config file type is being loaded correctly, for every possible config file type. I strongly suspect there will be an error or warning when loading an `mjs` file.
4. Any other tests that you can think of to prove that the underlying logic is both complete and correct. [We don't want the bar to explode when someone asks where the bathroom is](https://dev.to/this-is-learning/100-code-coverage-is-a-lie-1i1a)! 😁

package-lock.json
mcknasty: Thank you for pointing that out.  Very good feedback.  I think it's a good idea to migrate to lockfile version 3, but this change was unintentional.  To keep a lockfile version at 2, you essentially have to install packages with the following commands.  

```
$ nvm use default
Now using node v16.19.0 (npm v9.5.0)
$ nvm use 12.22.12
Now using node v12.22.12 (npm v8.19.3)
$ rm -rdf package-lock.json node_modules
$ npm i

added 366 packages, and audited 367 packages in 7s

89 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities
$ head package-lock.json
{
  "name": "c8",
  "version": "7.13.0",
  "lockfileVersion": 2,
  "requires": true,
  "packages": {
    "": {
      "name": "c8",
      "version": "7.13.0",
      "license": "ISC",
``` 

On a side note, they may have recently changed the version of npm with the LTS version of NodeJS 14. 

```
$ nvm use 14.21.3
Now using node v14.21.3 (npm v9.5.0)
```

lib/parse-args.js
mcknasty: This is a good suggestion.  

> While I'd love to see mjs support, the catch there is that ESM doesn't like getting loaded into a CommonJS process. The other way 'round can be done easily, but I think it's out of scope for this ticket to convert this package to ESM... :P

From my understanding, and from the comments above, we would like to remove the .mjs requirement as it was out of scope for this particular patch.  I am going to remove the '.mjs' from the ext list.

mcknasty: Based on the comment above, I think the acceptable list of file names should be the following in order of priority.

```
 .c8rc
.c8rc.json
.c8rc.yml
.c8rc.yaml
.c8.config.js
.c8.config.cjs
.nycrc
.nycrc.json
.nycrc.yml
.nycrc.yaml
nyc.config.js
nyc.config.cjs
```

test/fixtures/config/.nycrc.mjs
```diff
@@ -0,0 +1,11 @@
...
+  "reporter": [
+    "html",
+    "text"
+  ],
+  "lines": 95,
+  "branches": "82",
+  "statements": "95"
+}
+
+export default config;
```
mcknasty, line 11: Going to remove this file and add a patch at a later date for .mjs support

lib/parse-args.js
mcknasty: > Tests that prove that each config file type is being loaded correctly, for every possible config file type. I strongly suspect there will be an error or warning when loading an mjs file.

I have a strategy I would like to suggest to accomplish this goal and get some feedback.  In order to test each configuration file being loading, I would need to run c8 in a directory or subdirectory containing that configuration file.  IMO, I have two options

1.  I could add a folder to the test directory labeled config-test or something similar and manually add subdirectories for the various configuration files. 
2. I could add a before and after hook for each configuration file unit test, and create these subdirectories at run time.  Obviously, the after hook would clean up after the before hook.

 I think there are advantages and drawbacks to each of the options.  

For option 1, this is many files to add to the test directory and there could be possible duplications of files from the config directory.

For option 2, since the directory creation and tear down is at run time, this could potentially lead to false positive test results, files accidentally getting committed to the code base, if the after function doesn't execute correctly, and it can make it more difficult to read what the test case is attempting to accomplish.  

@bcoe and @kf6kjg Do you see any other approaches?  Is there a preference?

kf6kjg: Or you could use a dependency injection pattern where you can 
1. inject the found file path instead of reading the directory tree, and
2. inject the config you want to test in place of the file read.

This provides better and faster tests since there's no actual disk seeks and you are in 100% full control in each test.

One tool I've found that I like is [tsyringe](https://github.com/microsoft/tsyringe), but often for simple cases I'll just do something like 

```typescript
function loadConfigFile(path, ext, _di = { readJsFile: require, readFile: readFileSync }) {
  // same logic, just replaces the relevant blocks with things like
  config = JSON.parse(_di.readFile(path, 'utf8'));
}

// Or even more complex:

function loadConfigFile(path, ext, _di = { readJsFile: require, readYamlFile: (path) => require('js-yaml').load(readFileSync(path, 'utf8')), readJsonFile: (path) => JSON.parse(readFileSync(path, 'utf8')) }) {
  // same logic, just replaces the relevant blocks with things like
  config = _di.readJsonFile(path);
}
```

With that you can in your tests pass to the function all the overrides you need to prove that the function is behaving correctly - including proving that it is indeed calling the functions.

kf6kjg: As to the order and file names I prefer my tooling config files, even the JS-based ones, to be hidden files. Thus I promote the following ordered list:

```
.c8rc
.c8rc.json
.c8rc.yml
.c8rc.yaml
.c8rc.js
.c8rc.cjs
.c8.config.js
.c8.config.cjs
c8.config.js
c8.config.cjs
.nycrc
.nycrc.json
.nycrc.yml
.nycrc.yaml
nyc.config.js
nyc.config.cjs
```

Having an `rc` file be a JS file may seem odd to some, but it's a pattern that's supported in ESLint, Mocha, Commitlint, Lint-Staged, Prettier, and more.

README.md
```diff
@@ -18,13 +18,25 @@ The above example will output coverage metrics for `foo.js`.
...
 
+A configuration file can be specified by passing its path on the command line with `--config` or `-c`. If no config option is provided, c8 searches for files named in the table below starting from `cwd` and walking up the filesystem tree.
+
+A variety of file names are available in an effort to stay backwards compatible with nyc and ensure dynamic configuration options depending on the build system.
+
+| File name                                   | File Association   |
+|---------------------------------------------|--------------------|
+| `.c8rc`, `.c8rc.json`                       | JSON               |
+| `.c8rc.yml`, `.c8rc.yaml`                   | YAML               |
+| `.c8rc.js`, `.c8rc.cjs`                     | CommonJS export    |
```
koshic, line 33: .js may be CJS or  ESM, depends on 'type' field in package.json. please reflect it in that PR

koshic (changes requested):

lib/parse-args.js
bcoe: I agree, let's confirm or deny that this code is necessary.. If we're not hitting the line that sets `/tmp`, what is setting `/tmp`?

bcoe (commented):
@koshic thank you for the review on this.

@mcknasty apologies for the very slow reviews, I have had next to no time for open source these days, unfortunately (between my management day job, and my rock climbing hobby 😛 ).

I appreciate the work you've been putting in.

kf6kjg:
@mcknasty do you need/want some assistance bringing this to the finish line?

mcknasty:
> @mcknasty do you need/want some assistance bringing this to the finish line?

Thanks, @kf6kjg the follow-up.  I will take another look at this patch this week.

lib/parse-args.js
mcknasty: @bcoe @kf6kjg Just created an issue in the queue.  #489.  Let's move this particular discussion there.

lib/parse-args.js
mcknasty: I just created a list of supported files in a [function](https://github.com/bcoe/c8/pull/436/commits/6823fa2a1f7b25a9386f0b858357e31fc9e00dcc#diff-40d93e4e5f4b94fc8a34a87a3328651c24e855051bfcdd6944dee76f9c82b8d4R217-R238)

In terms of [test cases](https://github.com/bcoe/c8/pull/436#discussion_r1113740681), I created test cases supporting if the file was passed on the command line or detected as a supported config file.

Marking as resolved.

README.md
mcknasty: Add some additional documentation to the readme as suggested.

package-lock.json
mcknasty: I believe another patch changed the lock file version.  Removed my change.

kf6kjg (approved):
Looking good from just a read-over! I don't have bandwidth to do much more than just check it over visually, but I like what I see.

lib/parse-args.js
```diff
@@ -194,6 +194,49 @@ function buildYargs (withCommands = false) {
...
+  const ext = extname(path).toLowerCase()
+  if (jsExts.includes(ext)) {
+    config = require(path)
+  } else if (ymlExts.includes(ext)) {
+    config = require('js-yaml').load(readFileSync(path, 'utf8'))
+  } else if (ext === '.json' || fileName.slice(-2) === 'rc') {
+    config = JSON.parse(readFileSync(path, 'utf8'))
+  }
+
+  // Should the process die if none of these filename extensions are found?
```
koshic, line 212: we should try to require() it: if there is expected situation (user passed config.ts with ts-node loaded by --require in NODE_OPTIONS, or something similar) - fine, else there is an error and process will die (which is expected in that case).

README.md
```diff
@@ -18,13 +18,25 @@ The above example will output coverage metrics for `foo.js`.
...
-
-A configuration file can be specified by passing its path on the command line with `--config` or `-c`. If no config option is provided, c8 searches for files named `.c8rc`, `.c8rc.json`, `.nycrc`, or `.nycrc.json`, starting from
-`cwd` and walking up the filesystem tree.
+c8 can be configured via command-line flags, a `c8` section in `package.json`, or a configuration file on disk.
 
 When using `package.json` configuration or a dedicated configuration file, omit the `--` prefix from the long-form of the desired command-line option.
 
+A configuration file can be specified by passing its path on the command line with `--config` or `-c`. If no config option is provided, c8 searches for files named in the table below starting from `cwd` and walking up the filesystem tree.
+
+A variety of file names are available in an effort to stay backwards compatible with nyc and ensure dynamic configuration options depending on the build system.
```
koshic, line 27: backward compatibility is good thing, but what is the preferred way for new users? I propose to add some note about 'modern' configuration system (c8.config.js, right?) and 'legacy' (and may be 'deprecated') options.

README.md
```diff
@@ -18,13 +18,25 @@ The above example will output coverage metrics for `foo.js`.
...
+| File name                                                                                           | File Association   |
+|-----------------------------------------------------------------------------------------------------|--------------------|
+| `.c8rc`, `.c8rc.json`                                                                               | JSON               |
+| `.c8rc.yml`, `.c8rc.yaml`                                                                           | YAML               |
+| `.c8rc.js`, `.c8rc.cjs`, `.c8.config.js`, `.c8.config.cjs`, `c8.config.js`, `c8.config.cjs`         | CommonJS export*   |
+| `.nycrc`, `.nycrc.json`                                                                             | JSON               |
+| `.nycrc.yaml`, `.nycrc.yml`                                                                         | YAML               |
+| `.nycrc.js`, `.nycrc.cjs`, `nyc.config.js`, `nyc.config.cjs`, `.nyc.config.js`, `.nyc.config.cjs`   | CommonJS export*   |
+
+At the moment ESM syntax is not supported for writing c8 configuration files.  This may change in the future, but please note, C8 is written in CommonJS syntax.
```
koshic, line 38: can we add suggestion how to use ESM inside .cjs config for users who already migrated to ESM?

good example is https://eslint.org/docs/latest/use/configure/configuration-files-new
image

koshic (changes requested):

README.md
mcknasty: @koshic The short answer:
For this particular patch, ESM support wasnt required.  This was decided a few months ago.

The long answer:

I have a branch where I am working out the ESM support.  I wouldn't let it hold up anything you are working on at the moment.  At the time of working on this branch, yargs had limited support for async calls.  This can cause challenges with modules that need to be loaded async.  I am unclear if that block code you mentioned above will produce the desired results.  If you want to cut a branch and show proof of concept, I would be willing to take a look at it and compare it to my development branch.

koshic: @mcknasty sure, there is no ESM support from c8 side. I just propose to add a few lines for people who already use ESM in their projects - 'hey, we are still working on ESM support, but you can use CJS config via simple async IIFE and import your configs / tools via dynamic import'. 

"I am unclear if that block code you mentioned above will produce the desired results" - https://developer.mozilla.org/en-US/docs/Glossary/IIFE, is this a good enough proof?

mcknasty: @koshic No, the IIFE documentation page will not satisfy a change to the c8 documentation.  By proof of concept, I am referring to a working example in a c8 fork or a public repo.  Again, due to the fact that yargs isn't configured for async calls, the configuration you mentioned above may not work.

For those using ESM modules, I would suggest using a cjs file with CommonJS, a Yaml file, or a JSON file for c8 configuration.

koshic: @mcknasty ok, let's stop this discussion - we're just wasting time on a temporary solution that will be eliminated by your next PR. thank you for your patience!

lib/parse-args.js
mcknasty: @koshic added an error to throw if an invalid configuration file is passed with the config arguments

README.md
mcknasty: I changed the wording to clarify what is supported for configuration options.  It includes currently supported ESM module options.

README.md
mcknasty: @koshic Thank you very much for your review.  Playing devil's advocate is very much a part of community open-source programming.  More reviews of any patch greatly improve the quality of any code base.  

If you could review the other items you mentioned for this patch, I would appreciate it.  Again thank you for your notes and contributions.

lib/parse-args.js
koshic: So, TypeScript users can't provide their absolute valid CJS configs just because... just because we decided to don't allow that? 

Again, 'it's ok because it's better than nothing' (c) and I really appreciate your efforts, but from user perspective it looks like unreasonable restrictions. Good example is Yarn and constraints configuration file - it should be named only as 'yarn.config.js', AND must be CJS. Already migrated to ESM? Sorry. TypeScript user? Sorry.

koshic (approved):

lib/parse-args.js
```diff
@@ -194,6 +194,52 @@ function buildYargs (withCommands = false) {
...
+    'c8.config.js',
+    'c8.config.cjs',
+    '.nycrc',
+    '.nycrc.json',
+    '.nycrc.yml',
+    '.nycrc.yaml',
+    '.nyc.config.js',
+    '.nyc.config.cjs',
+    'nyc.config.js',
+    'nyc.config.cjs'
```
mcknasty, line 239: Should I include .c8rc.json, c8rc.json, .c8.config.json, c8.config.json, .nyc.config.json, and nyc.config.json as valid configuration names?

kf6kjg: I'd say that expanding scope like that should be the subject of an additional PR. The scope here is mainly adding YAML and CJS config support.

mcknasty:
A few things.

1.  I am seeing a large drop in code coverage, even though the patch has not changed much.
2.  I also see the failure of unit tests in versions 18.x and 20.x of node.js. 

Oddly enough, on commit [a9b6726d0d99769bc0ce886afdc95b5b4a999a11](https://github.com/bcoe/c8/pull/486/commits/a9b6726d0d99769bc0ce886afdc95b5b4a999a11), I am not seeing the code coverage drops, but I am seeing the test case failures in for the 18.x and 20.x branches of node.

I am looking into the code coverage drops concern a bit more.

test/parse-args.js
```diff
@@ -39,6 +67,87 @@ describe('parse-args', () => {
...
+          'nyc.config.cjs': 94,
+          '.nyc.config.js': 85,
+          '.nyc.config.cjs': 71
+        }
+
+        testReadingConfigFile(fileNameLineNumberMap, filePath)
+
+        afterEach(afterTestReadingConfigFile)
+      })
+    })
```
mcknasty, line 149: If I comment out this block of code, code coverage improves back to acceptable levels.

Current Coverage if this block is included.

```#/bin/bash
All files                 |   75.67 |     67.1 |   71.87 |   75.67 |                                
 bin                      |   86.04 |       60 |     100 |   86.04 |                                
  c8.js                   |   86.04 |       60 |     100 |   86.04 | 20,33-35,41-42                 
 lib                      |   79.97 |    69.23 |      84 |   79.97 |                                
  parse-args.js           |   98.52 |     90.9 |     100 |   98.52 | 165-166,187-188                
  report.js               |   65.42 |       60 |   71.42 |   65.42 | ...354,358-359,375-380,391-392 
  source-map-from-file.js |      88 |    53.84 |     100 |      88 | 48-49,59,74-76,91-93,95-97     
 lib/commands             |   41.96 |       50 |   16.66 |   41.96 |                                
  check-coverage.js       |   18.57 |      100 |       0 |   18.57 | 9-11,14-36,39-53,55-70         
  report.js               |   80.95 |       40 |      50 |   80.95 | 9-10,15-20                     
ERROR: Coverage for lines (75.67%) does not meet global threshold (95%)
ERROR: Coverage for branches (67.1%) does not meet global threshold (82%)
ERROR: Coverage for statements (75.67%) does not meet global threshold (95%)
Error: Process completed with exit code 1.
```

test/parse-args-helper.js
```diff
@@ -0,0 +1,55 @@
...
+      // set the snapshot filename
+      chaiJestSnapshot.setFilename('./test/fixtures/config/snapshots/' + fileName + '.snap')
+
+      copyFileSync('./test/fixtures/config/' + fileName, testPath + '/' + fileName)
+      // Run V8 in the dir above
+      const { output } = spawnSync(nodePath, [
+        c8Path,
+        nodePath,
+        require.resolve('./fixtures/tmp-config-test/normal.js')
+      ])
```
mcknasty: These lines are causing false reporting of uncovered lines.

mcknasty:
# Overview 
:clap: The Good - This Pull Request is getting close.
:ok_hand: The Bad - Need to get another Pull Request merged before this one is ready to go. More details on that in the next few days.
:facepunch: The Ugly - One of the Workflows is causing errors on multiple pull requests.

I have the below error in the workflow that I am unable to reproduce locally.  I am seeing the [same error](https://github.com/bcoe/c8/actions/runs/7522210164/job/20473934533?pr=453#step:4:1) in (#453) and [again]() in (#452) 

## Workflow Errors

commit id checkout: [27d619c1b6005d0563098ab2d9909117586aa852](https://github.com/bcoe/c8/actions/runs/7524917900/job/20480438662?pr=436#step:2:77)
`npm ci --engine-strict` [command failing](https://github.com/bcoe/c8/actions/runs/7524917900/job/20480438662?pr=436#step:4:1)
```#!/bin/bash
Run npm ci --engine-strict
  npm ci --engine-strict
  shell: /usr/bin/bash -e {0}
npm ERR! Cannot read property '@bcoe/v8-coverage' of undefined

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/runner/.npm/_logs/2024-01-15T05_55_22_507Z-debug.log
Error: Process completed with exit code 1.
```

## Attempt to Reproduce Locally
```#!/bin/bash
$ git remote -v
origin  git@github.com:mcknasty/c8.git (fetch)
origin  git@github.com:mcknasty/c8.git (push)
upstream        https://github.com/bcoe/c8.git (fetch)
upstream        https://github.com/bcoe/c8.git (push)
$ git branch | grep '*'
* additional-config-files
$ git log | head -n 1
commit 27d619c1b6005d0563098ab2d9909117586aa852
$ node -v
v14.21.3
$ npm -v
9.8.1
$ npm ci --engine-strict

added 383 packages, and audited 384 packages in 5s

98 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities
$ uname --all
Linux XXXXXXXXX 5.15.133.1-microsoft-standard-WSL2 #1 SMP Thu Oct 5 21:02:42 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
```

Is there a way to run the github actions locally?  I initially tried a [package called act](https://github.com/nektos/act).  I didn't spend a ton of time with it, but it seemed to have some challenges.  Are there any other packages I can try to reproduce this error or some additional information on `act` that could reproduce the errors in the workflow?

test/parse-args-helper.js
```diff
@@ -0,0 +1,61 @@
...
+          '--all',
+          '--src=./test/fixtures/tmp-config-test',
+          nodePath,
+          require.resolve('./fixtures/tmp-config-test/normal.js')
+        ],
+        { cwd: './test/fixtures/tmp-config-test' }
+      )
+      output.toString('utf8').should.matchSnapshot()
+    })
+  })
```
mcknasty, line 39: Last nitpick 🐞

mcknasty:
Skipped the config detection tests for the moment.  Ready to go from my perspective.

lib/parse-args.js
```diff
@@ -194,6 +194,52 @@ function buildYargs (withCommands = false) {
...
+    config = require(path)
+  } else if (ymlExts.includes(ext)) {
+    config = require('js-yaml').load(readFileSync(path, 'utf8'))
+  } else if (ext === '.json' || fileName.slice(-2) === 'rc') {
+    config = JSON.parse(readFileSync(path, 'utf8'))
+  }
+
+  // Should the process die if none of these filename extensions are found?
+  if (Object.keys(config).length === 0) {
+    throw new Error(`Unsupported file type ${ext} while reading file ${path}`)
```
kf6kjg, line 214: I read why this was added, but it's going to result in bad UX when a config file return an empty object - in that situation it loaded fine, and is not an invalid file type, it's just not got any config entries.

lib/parse-args.js
```diff
@@ -194,6 +194,52 @@ function buildYargs (withCommands = false) {
...
+  const fileName = basename(path)
+
+  const ext = extname(path).toLowerCase()
+  if (jsExts.includes(ext)) {
+    config = require(path)
+  } else if (ymlExts.includes(ext)) {
+    config = require('js-yaml').load(readFileSync(path, 'utf8'))
+  } else if (ext === '.json' || fileName.slice(-2) === 'rc') {
+    config = JSON.parse(readFileSync(path, 'utf8'))
+  }
```
mcknasty, line 210: I am going to put this into a try/catch block for the following reasons

1.  If a not properly formated json or yaml object is returned it will produce an error.
2. If the path doesn't exist, it will produce an error.
3.  It could pass an invalid configuration key is pass, it will essentially get ignored.

It could in the future be helpful to have a config schema validator.  Going to add unit tests for two of the potential issues above.

lib/parse-args.js
mcknasty: I want to be sure I am clear on what you stated above.  To rewrite what you expressed in two use cases.

Scenario:  

A user executes the c8 command with the --config parameter set to an existing file on the file system

Given:

  The file has a valid file name as documented in the c8 readme
  The file has no contents.

Scenario:  

A user executes the c8 command without additional configuration arguments and c8 discovers a configuration file.

Given:
   The file has a valid file name as documented in the c8 readme
   The file has no contents.


Can you please confirm?

The line does cover the following use case:

Scenario:
  A user executes the c8 command with the --config parameter set to an existing file on the file system.

Given:
  The file does not have a valid file name as documented in the c8 readme

Result:
   c8 will produce the error `Unsupported file type ...'

This test case can be found on [Line 70 - Line 85](https://github.com/bcoe/c8/blob/22a9c08a01a219a97e9768be7b211a6793c22956/test/parse-args.js#L70-L85) on the file test/parse-args.js

test/parse-args-helper.js
```diff
@@ -0,0 +1,63 @@
...
+          '--temp-directory=tmp/normal',
+          '--all',
+          '--src=./test/fixtures/tmp-config-test',
+          nodePath,
+          require.resolve('./fixtures/tmp-config-test/normal.js')
+        ],
+        { cwd: './test/fixtures/tmp-config-test' }
+      )
+      output.toString('utf8').should.matchSnapshot()
+    })
```
mcknasty, line 40: Move this logic to test/parse-args.js. Change file name to helpers.js and refactor `testReadingConfigFile` to add a parameter for callback function that contains test cases.

lib/parse-args.js
kf6kjg: Some notes before I hit the scenarios:

1. Only YAML would have an empty file return an empty object.
2. An empty JSON file is a syntax error.
3. A JSON file of `{}` is valid.
4. A CJS script that doesn't export anything IMO SHOULD result in an error as this is a common mistake.
5. A CJS script that exports an empty object, e.g. `{}` is valid. 

So, scenarios:

----
Scenario:
A user executes the c8 command with the --config parameter set to an existing file on the file system
OR
A user executes the c8 command without the --config parameter set AND c8 discovers a configuration file.

Given any of:
- The file is JSON with the `.json` extension, case insensitive.
  The file has the contents `{}` or equivalent.
- The file is YAML with the `.yaml` or `.yml` extension, case insensitive.
  The file has no contents
- The file is CommonJS with the `.js` or `.cjs` extension, case insensitive.
  The file's contents result in an empty object `{}` being exported.

Result:
c8 will run with a default configuration.

----
Scenario:
A user executes the c8 command with the --config parameter set to an existing file on the file system
OR
A user executes the c8 command without the --config parameter set AND c8 discovers a configuration file.

Given any of:
- The file has the `.json` extension, case insensitive.
  The file is invalid JSON.
- The file has the `.yaml` or `.yml` extension, case insensitive.
  The file is invalid YAML.
- The file has the `.js` or `.cjs` extension, case insensitive.
  The file's contents are invalid CommonJS or have a runtime error.
- The file has the `.js` or `.cjs` extension, case insensitive.
  The file doesn't export anything.
- The file has the `.js` or `.cjs` extension, case insensitive.
  The file exports something that's not an object.

Result:
c8 will fail with an appropriate error based on the actual underlying problem that will be helpful to the user.

Some errors are really very obtuse or otherwise mask the real problem. I think over time of people testing this those errors will be discovered and made nicer, so making the errors nicer probably shouldn't be part of this PR.

----
Scenario:
A user executes the c8 command with the --config parameter set to an existing file on the file system.

Given:
The file name is NOT one of the ones documented in the c8 readme.
The file extension is one of the valid types, e.g. `.js, .cjs, .json, .yaml, .yml` etc.
The file returns a valid c8 configuration.

Result:
c8 will run with the configuration coming from the file given.

This is because there are several reasons for using the `--config` parameter, here's just a few:
1. I've got multiple config files to choose from. This you've covered.
2. My config file's at an unusual path that auto detection won't find, e.g. `.configs/c8rc.yaml` This I'm pretty sure you've covered, though maybe not directly with a test.
3. I don't like any of the default file names and have my own naming scheme I use, e.g. `.config.c8.yaml` - the current code looks like it handles this case already, there just doesn't seem to be a test covering this idea.

lib/parse-args.js
kf6kjg: I agree on a schema validator. It would help people a lot.

However masking erroneous configuration files by just ignoring them, or even just logging the problem and proceeding with defaults, is an antipattern to me. It results in unexpected behavior since c8 proceeds with different settings than intended.

lib/parse-args.js
kf6kjg: I've [created a PR](https://github.com/mcknasty/c8/pull/5) for the branch of this PR that adds tests and changes to support the above.

lib/parse-args.js
kf6kjg: In my [PR to this PR](https://github.com/mcknasty/c8/pull/5) I did wrap some of these in try-catch, but I only caught specific errors, re-thowing the others. The idea was to enhance the message in those cases as the underlying error was obtuse. However I think that other errors, such as file not found, are generally acceptable to leave alone.

kf6kjg:
Wait what, why closed?😳

mcknasty:
sorry was editing my name in the commit history and it closed the branch.   I'll fix it.

mcknasty:
Still won't let me reopen due to a force push.

kf6kjg:
For future visitors, including myself, the replacement PR is https://github.com/bcoe/c8/pull/518

PR #447, opened 2023-01-17
Increase Code Quality By Increasing Code Coverage: lib/parse-args.js

mcknasty:
# **[Pull Request 447](https://github.com/bcoe/c8/pull/447)**

Increase code quality by increasing code coverage: lib/parse-args.js

refactor: bumping code coverage to 100% for lib/parse-args.js. (#447)
test: Adding a test case for NODE_V8_COVERAGE and changing the describe block title. (#447)
test: Early exit branch test for hideInstrumenteeArgs function. (#447)
test: Adding test for relative paths for relative report directory cli option. (#447)
test: Adding test case for relative paths for temporary directory cli option. (#447)
test: Adding test for passing node.js arguments to c8. (#447)

## **Following the Contributions Recommendations [here.](https://github.com/bcoe/c8/blob/main/CONTRIBUTING.md)**

1. [x] Make sure you have installed the latest version of Node.js
2. [x] Fork this project on GitHub and clone your fork locally
3. [x] Create local branches to work within. These should also be created directly from the main branch.  Local Fork [here.](https://github.com/mcknasty/c8)
4. [x] Make your changes.
5. [x] Run tests to make sure all is okay (everything should pass except the snapshot).
   1. A [complete log](https://gist.github.com/mcknasty/9cb0f0ed7c6c63210fd6f345e4796166#file-initial-test-txt) of initial test results.
   2. As instructed, ignore snapshot failures. 0 failures
   3. [100 passing](https://gist.github.com/mcknasty/9cb0f0ed7c6c63210fd6f345e4796166#file-initial-test-txt-L151) in [48 seconds](https://gist.github.com/mcknasty/9cb0f0ed7c6c63210fd6f345e4796166#file-initial-test-txt-L151)
6. [x] Now update the snapshot.
   1. A [complete log](https://gist.github.com/mcknasty/9cb0f0ed7c6c63210fd6f345e4796166#file-snapshot-test-txt) of snapshot test results.
   2. [100 passing](https://gist.github.com/mcknasty/9cb0f0ed7c6c63210fd6f345e4796166#file-snapshot-test-txt-L155) in [48 seconds](https://gist.github.com/mcknasty/9cb0f0ed7c6c63210fd6f345e4796166#file-snapshot-test-txt-L155)
7. [x] If all is passing, commit your changes. The log of the commit can be found [here.](https://github.com/bcoe/c8/pull/447/commits/7e611964a10968425f643a45124db1b0beabd4f5)
8. [x] As a best practice, once you have committed your changes, it is a good idea to use git rebase (not git merge) to synchronize your work with the main repository.
9. [x] Run tests again to make sure all is okay.
    1. A [complete log](https://gist.github.com/mcknasty/9cb0f0ed7c6c63210fd6f345e4796166#file-final-test-txt) of the final test results.
    2. [100 passing](https://gist.github.com/mcknasty/9cb0f0ed7c6c63210fd6f345e4796166#file-final-test-txt-L151) in [48 seconds](https://gist.github.com/mcknasty/9cb0f0ed7c6c63210fd6f345e4796166#file-final-test-txt-L151)
10. [x] Push
11. [x] Open the pull request, see details in the template.
12. [ ] Make any necessary changes after review.

## **File Code Coverage Matrix Report**

|   Test Type   | File    | Statement | Branch | Function | Lines |
|---------------|---------|-----------|--------|----------|-------|
| Final Test | [lib/parse-args.js](https://gist.github.com/mcknasty/9cb0f0ed7c6c63210fd6f345e4796166#file-final-test-txt-L160) | 100% | 100% | 100% | 100% |
| Node 14.21.3 | [lib/parse-args.js](https://gist.github.com/mcknasty/858aa5d8ea51fba08ef0d2062632180d#file-v14-21-3-test-txt-L183) | 100% | 100% | 100% | 100% |
| Node 16.20.2 | [lib/parse-args.js](https://gist.github.com/mcknasty/858aa5d8ea51fba08ef0d2062632180d#file-v16-20-2-test-txt-L184) | 100% | 100% | 100% | 100% |
| Node 18.19.0 | [lib/parse-args.js](https://gist.github.com/mcknasty/858aa5d8ea51fba08ef0d2062632180d#file-v18-19-0-test-txt-L180) | 100% | 100% | 100% | 100% |
| Node 20.11.0 | [lib/parse-args.js](https://gist.github.com/mcknasty/858aa5d8ea51fba08ef0d2062632180d#file-v20-11-0-test-txt-L180) | 100% | 100% | 100% | 100% |

## **Unit Tests Results**

All tests run on Node version 18.19.0

|   Test Type     |  PASS          |  Tests Passed   |  Tests Failed  |    Time      |
|:---------------:|:--------------:|:---------------:|:--------------:|:------------:|
| [Initial Test](https://gist.github.com/mcknasty/9cb0f0ed7c6c63210fd6f345e4796166#file-initial-test-txt)       | :heavy_check_mark: | [100 passing](https://gist.github.com/mcknasty/9cb0f0ed7c6c63210fd6f345e4796166#file-initial-test-txt-L151)        | 0 failures       | [48 seconds](https://gist.github.com/mcknasty/9cb0f0ed7c6c63210fd6f345e4796166#file-initial-test-txt-L151)       |
| [Snapshot Test](https://gist.github.com/mcknasty/9cb0f0ed7c6c63210fd6f345e4796166#file-snapshot-test-txt)       | :heavy_check_mark: | [100 passing](https://gist.github.com/mcknasty/9cb0f0ed7c6c63210fd6f345e4796166#file-snapshot-test-txt-L155)        | 0 failures       | [48 seconds](https://gist.github.com/mcknasty/9cb0f0ed7c6c63210fd6f345e4796166#file-snapshot-test-txt-L155)       |
| [Final Test](https://gist.github.com/mcknasty/9cb0f0ed7c6c63210fd6f345e4796166#file-final-test-txt)       | :heavy_check_mark: | [100 passing](https://gist.github.com/mcknasty/9cb0f0ed7c6c63210fd6f345e4796166#file-final-test-txt-L151)        | 0 failures       | [48 seconds](https://gist.github.com/mcknasty/9cb0f0ed7c6c63210fd6f345e4796166#file-final-test-txt-L151)       |

## **Node Version Testing Matrix**

|   Node Version  |  PASS          |  Tests Passed   |  Tests Failed  |    Time      |
|:---------------:|:--------------:|:---------------:|:--------------:|:------------:|
| [14.21.3](https://gist.github.com/mcknasty/858aa5d8ea51fba08ef0d2062632180d#file-v14-21-3-test-txt)       | :heavy_check_mark: | [100 passing](https://gist.github.com/mcknasty/858aa5d8ea51fba08ef0d2062632180d#file-v14-21-3-test-txt-L174)        | 0 failures       | [55 seconds](https://gist.github.com/mcknasty/858aa5d8ea51fba08ef0d2062632180d#file-v14-21-3-test-txt-L174)       |
| [16.20.2](https://gist.github.com/mcknasty/858aa5d8ea51fba08ef0d2062632180d#file-v16-20-2-test-txt)       | :heavy_check_mark: | [100 passing](https://gist.github.com/mcknasty/858aa5d8ea51fba08ef0d2062632180d#file-v16-20-2-test-txt-L175)        | 0 failures       | [1 minutes](https://gist.github.com/mcknasty/858aa5d8ea51fba08ef0d2062632180d#file-v16-20-2-test-txt-L175)       |
| [18.19.0](https://gist.github.com/mcknasty/858aa5d8ea51fba08ef0d2062632180d#file-v18-19-0-test-txt)       | :heavy_check_mark: | [100 passing](https://gist.github.com/mcknasty/858aa5d8ea51fba08ef0d2062632180d#file-v18-19-0-test-txt-L171)        | 0 failures       | [55 seconds](https://gist.github.com/mcknasty/858aa5d8ea51fba08ef0d2062632180d#file-v18-19-0-test-txt-L171)       |
| [20.11.0](https://gist.github.com/mcknasty/858aa5d8ea51fba08ef0d2062632180d#file-v20-11-0-test-txt)       | :heavy_check_mark: | [100 passing](https://gist.github.com/mcknasty/858aa5d8ea51fba08ef0d2062632180d#file-v20-11-0-test-txt-L171)        | 0 failures       | [1 minutes](https://gist.github.com/mcknasty/858aa5d8ea51fba08ef0d2062632180d#file-v20-11-0-test-txt-L171)       |

Referencing Issue #448

Updated 2023-07-29 7:59 PM EST
Updated 2023-07-30 11:25 AM EST
Updated 2024-01-14 9:03 PM EST

### DIFF
--- a/lib/parse-args.js
+++ b/lib/parse-args.js
@@ -30,7 +30,8 @@ function buildYargs (withCommands = false) {
       alias: ['o', 'report-dir'],
       group: 'Reporting options',
       describe: 'directory where coverage reports will be output to',
-      default: './coverage'
+      default: './coverage',
+      type: 'string'
     })
     .options('all', {
       default: false,
@@ -126,7 +127,8 @@ function buildYargs (withCommands = false) {
     })
     .option('temp-directory', {
       describe: 'directory V8 coverage data is written to and read from',
-      default: process.env.NODE_V8_COVERAGE
+      default: process.env.NODE_V8_COVERAGE,
+      type: 'string'
     })
     .option('clean', {
       default: true,
@@ -207,12 +209,12 @@ function hideInstrumenteeArgs () {
   let argv = process.argv.slice(2)
   const yargv = parser(argv)
 
-  if (!yargv._.length) return argv
-
-  // drop all the arguments after the bin being
-  // instrumented by c8.
-  argv = argv.slice(0, argv.indexOf(yargv._[0]))
-  argv.push(yargv._[0])
+  if (yargv._.length !== 0) {
+    // drop all the arguments after the bin being
+    // instrumented by c8.
+    argv = argv.slice(0, argv.indexOf(yargv._[0]))
+    argv.push(yargv._[0])
+  }
 
   return argv
 }
--- a/test/parse-args.js
+++ b/test/parse-args.js
@@ -15,6 +15,11 @@ describe('parse-args', () => {
       const instrumenterArgs = hideInstrumenteeArgs()
       instrumenterArgs.should.eql(['--foo=99', 'my-app'])
     })
+    it('test early exit from function if no arguments are passed', () => {
+      process.argv = []
+      const instrumenterArgs = hideInstrumenteeArgs()
+      instrumenterArgs.length.should.eql(0)
+    })
   })
 
   describe('hideInstrumenterArgs', () => {
@@ -25,6 +30,13 @@ describe('parse-args', () => {
       instrumenteeArgs.should.eql(['my-app', '--help'])
       argv.tempDirectory.endsWith(join('coverage', 'tmp')).should.be.equal(true)
     })
+    it('interprets first args after -- as Node.js execArgv', () => {
+      const expected = [process.execPath, '--expose-gc', 'index.js']
+      process.argv = ['node', 'c8', '--', '--expose-gc', 'index.js']
+      const argv = buildYargs().parse(hideInstrumenteeArgs())
+      const munged = hideInstrumenterArgs(argv)
+      munged.should.deep.equal(expected)
+    })
   })
 
   describe('with NODE_V8_COVERAGE already set', () => {
@@ -34,8 +46,20 @@ describe('parse-args', () => {
       process.argv = ['node', 'c8', '--foo=99', 'my-app', '--help']
       const argv = buildYargs().parse(hideInstrumenteeArgs())
       argv.tempDirectory.endsWith('/coverage/tmp_').should.be.equal(true)
+      delete process.env.NODE_V8_COVERAGE
       process.env.NODE_V8_COVERAGE = NODE_V8_COVERAGE
     })
+    /** Commenting out till the print config PR is ready * /
+    it('should set it if undefined', () => {
+      const NODE_V8_COVERAGE = process.env.NODE_V8_COVERAGE
+      process.env.NODE_V8_COVERAGE = undefined
+      process.argv = ['node', 'c8', '--foo=99', 'my-app', '--help']
+      const argv = buildYargs().parse(hideInstrumenteeArgs())
+      argv.tempDirectory.endsWith('/coverage/tmp').should.be.equal(true)
+      delete process.env.NODE_V8_COVERAGE
+      process.env.NODE_V8_COVERAGE = NODE_V8_COVERAGE
+    })
+    /** */
   })
 
   describe('--config', () => {
--- a/test/integration.js.snap
+++ b/test/integration.js.snap
@@ -166,7 +166,7 @@ All files                              |    3.52 |     12.5 |    6.52 |    3.52 
   prettify.js                          |       0 |        0 |       0 |       0 | 1-2                    
   sorter.js                            |       0 |        0 |       0 |       0 | 1-196                  
  c8/lib                                |       0 |        0 |       0 |       0 |                        
-  parse-args.js                        |       0 |        0 |       0 |       0 | 1-224                  
+  parse-args.js                        |       0 |        0 |       0 |       0 | 1-226                  
   report.js                            |       0 |        0 |       0 |       0 | 1-402                  
   source-map-from-file.js              |       0 |        0 |       0 |       0 | 1-100                  
  c8/lib/commands                       |       0 |        0 |       0 |       0 |                        
@@ -531,7 +531,7 @@ All files                              |    3.52 |     12.5 |    6.52 |    3.52 
   prettify.js                          |       0 |        0 |       0 |       0 | 1-2                    
   sorter.js                            |       0 |        0 |       0 |       0 | 1-196                  
  c8/lib                                |       0 |        0 |       0 |       0 |                        
-  parse-args.js                        |       0 |        0 |       0 |       0 | 1-224                  
+  parse-args.js                        |       0 |        0 |       0 |       0 | 1-226                  
   report.js                            |       0 |        0 |       0 |       0 | 1-402                  
   source-map-from-file.js              |       0 |        0 |       0 |       0 | 1-100                  
  c8/lib/commands                       |       0 |        0 |       0 |       0 |                        


bcoe (changes requested):
Thanks for the contribution, I'm excited to improve the coverage of c8.

In this PR specifically, I don't like that it changes the behaviour in subtle ways:

* default will no longer give a hint in the UI that shows that we've defaulted to NODE_V8_COVERAGE.
* The path of `tempDirectory` will no longer be an absolute path (`./tmp`, vs., `/full/path/to/tmp`).

I think you should be able to get coverage up without making these changes, one option would be just adding some ignore lines if necessary (since the logic is part of the setup of yargs, and not really part of the application logic).

lib/parse-args.js
```diff
@@ -5,9 +5,12 @@ const { readFileSync } = require('fs')
 const Yargs = require('yargs/yargs')
 const { applyExtends } = require('yargs/helpers')
 const parser = require('yargs-parser')
-const { resolve } = require('path')
 
 function buildYargs (withCommands = false) {
+  if (typeof process.env.NODE_V8_COVERAGE !== 'string') {
```
mcknasty, line 10: Was able to reduce this logic down to a single line

mcknasty, line 10: Was able to refactor the block into three lines.  The relative path has been changed to an absolute path.

test/parse-args.js
```diff
@@ -36,6 +36,14 @@ describe('parse-args', () => {
       argv.tempDirectory.endsWith('/coverage/tmp_').should.be.equal(true)
       process.env.NODE_V8_COVERAGE = NODE_V8_COVERAGE
     })
+    it('should set it if undefined', () => {
```
mcknasty, line 39: Added an additional Unit Test

lib/parse-args.js
```diff
@@ -126,7 +129,8 @@ function buildYargs (withCommands = false) {
     })
     .option('temp-directory', {
       describe: 'directory V8 coverage data is written to and read from',
-      default: process.env.NODE_V8_COVERAGE
+      default: process.env.NODE_V8_COVERAGE,
```
mcknasty, line 132: I changed this line back to what it was in earlier versions.  This should address the UI prompts as indicated in the [PR comments.](https://github.com/bcoe/c8/pull/447#pullrequestreview-1265881416)

mcknasty (commented):
@bcoe Hopefully this new patch will address the concerns in your comments.  Thank you for taking a look at this.

lib/parse-args.js
```diff
@@ -154,12 +158,6 @@ function buildYargs (withCommands = false) {
     })
     .pkgConf('c8')
     .demandCommand(1)
-    .check((argv) => {
```
mcknasty, line 157: From my understanding, this function should not actually change arguments but only checks to see if they are valid.

lib/parse-args.js
```diff
@@ -191,18 +189,13 @@ function buildYargs (withCommands = false) {
 function hideInstrumenterArgs (yargv) {
   let argv = process.argv.slice(1)
   argv = argv.slice(argv.indexOf(yargv._[0]))
-  if (argv[0][0] === '-') {
```
mcknasty, line 194: Not sure what use case this was addressing.  From what I can tell, if the first character of the first argument is a `-` then add the path of where C8 is located to the end of the argument array.  What's the purpose of this?  Do I need to add a Test Case instead of removing the code.

lib/parse-args.js
```diff
@@ -191,18 +189,13 @@ function buildYargs (withCommands = false) {
...
-    argv.unshift(process.execPath)
-  }
   return argv
 }
 
 function hideInstrumenteeArgs () {
   let argv = process.argv.slice(2)
   const yargv = parser(argv)
 
-  if (!yargv._.length) return argv
```
mcknasty, line 204: `yargv._` is this a call to underscore?  Is there another Test Case I need to add for this condition?

lib/parse-args.js
```diff
@@ -154,12 +158,6 @@ function buildYargs (withCommands = false) {
     })
     .pkgConf('c8')
     .demandCommand(1)
-    .check((argv) => {
-      if (!argv.tempDirectory) {
-        argv.tempDirectory = resolve(argv.reportsDir, 'tmp')
-      }
-      return true
-    })
```
mcknasty, line 162: `From` my understanding, this function should not actually change arguments but only checks to see if they are valid.

lib/parse-args.js
```diff
@@ -126,7 +129,8 @@ function buildYargs (withCommands = false) {
     })
     .option('temp-directory', {
       describe: 'directory V8 coverage data is written to and read from',
-      default: process.env.NODE_V8_COVERAGE
+      default: process.env.NODE_V8_COVERAGE,
+      type: 'string'
     })
```
mcknasty, line 134: I changed this line back to what it was in earlier versions.  This should address the UI prompts as indicated in the [PR comments.](https://github.com/bcoe/c8/pull/447#pullrequestreview-1265881416)

lib/parse-args.js
```diff
@@ -5,9 +5,12 @@ const { readFileSync } = require('fs')
 const Yargs = require('yargs/yargs')
 const { applyExtends } = require('yargs/helpers')
 const parser = require('yargs-parser')
-const { resolve } = require('path')
 
 function buildYargs (withCommands = false) {
+  if (typeof process.env.NODE_V8_COVERAGE !== 'string') {
+    process.env.NODE_V8_COVERAGE = process.cwd() + '/coverage/tmp'
+  }
+
```
mcknasty, line 13: Was able to refactor the block into three lines.  The relative path has been changed to an absolute path.

lib/parse-args.js
```diff
@@ -191,18 +189,13 @@ function buildYargs (withCommands = false) {
 function hideInstrumenterArgs (yargv) {
   let argv = process.argv.slice(1)
   argv = argv.slice(argv.indexOf(yargv._[0]))
-  if (argv[0][0] === '-') {
-    argv.unshift(process.execPath)
-  }
   return argv
 }
```
mcknasty, line 193: This condition was never reached per the former code coverage report.

What use case does this line address?  I didn't find any documentation on passing a `-` for options via the command line.  

From what I can tell, if the first character of the second argument is a `-` then add the path of where C8 is located to the end of the argument array.  What purpose does this serve?  Do I need to add a Test Case instead of removing the code?

lib/parse-args.js
```diff
@@ -191,18 +189,13 @@ function buildYargs (withCommands = false) {
...
 
 function hideInstrumenteeArgs () {
   let argv = process.argv.slice(2)
   const yargv = parser(argv)
 
-  if (!yargv._.length) return argv
-
   // drop all the arguments after the bin being
   // instrumented by c8.
   argv = argv.slice(0, argv.indexOf(yargv._[0]))
```
mcknasty, line 201: `yargv._` is this a call to underscore?  Is there another Test Case I need to add for this condition?  This condition was never reached per the former code coverage report.

test/parse-args.js
```diff
@@ -36,6 +36,14 @@ describe('parse-args', () => {
...
+      const NODE_V8_COVERAGE = process.env.NODE_V8_COVERAGE
+      delete process.env.NODE_V8_COVERAGE
+      process.argv = ['node', 'c8', '--foo=99', 'my-app', '--help']
+      const argv = buildYargs().parse(hideInstrumenteeArgs())
+      argv.tempDirectory.endsWith('/coverage/tmp').should.be.equal(true)
+      process.env.NODE_V8_COVERAGE = NODE_V8_COVERAGE
+    })
   })
```
mcknasty, line 47: Added an additional unit test and changed the title of the describe block.

lib/parse-args.js
```diff
@@ -5,9 +5,11 @@ const { readFileSync } = require('fs')
...
 const { applyExtends } = require('yargs/helpers')
 const parser = require('yargs-parser')
-const { resolve } = require('path')
 
 function buildYargs (withCommands = false) {
+  if (typeof process.env.NODE_V8_COVERAGE !== 'string') {
+    process.env.NODE_V8_COVERAGE = process.cwd() + '/coverage/tmp'
+  }
   const yargs = Yargs([])
     .usage('$0 [opts] [script] [opts]')
```
mcknasty, line 14: Was able to refactor the block into three lines.  The relative path has been changed to an absolute path.

mcknasty, line 14: Was able to refactor the block into three lines.  The relative path has been changed to an absolute path.

lib/parse-args.js
```diff
@@ -126,7 +128,8 @@ function buildYargs (withCommands = false) {
     })
     .option('temp-directory', {
       describe: 'directory V8 coverage data is written to and read from',
-      default: process.env.NODE_V8_COVERAGE
+      default: process.env.NODE_V8_COVERAGE,
+      type: 'string'
     })
```
mcknasty, line 133: I changed this line back to what it was in earlier versions.  This should address the UI prompts as indicated in the [PR comments.](https://github.com/bcoe/c8/pull/447#pullrequestreview-1265881416)

lib/parse-args.js
```diff
@@ -154,12 +157,6 @@ function buildYargs (withCommands = false) {
...
     .pkgConf('c8')
     .demandCommand(1)
-    .check((argv) => {
-      if (!argv.tempDirectory) {
-        argv.tempDirectory = resolve(argv.reportsDir, 'tmp')
-      }
-      return true
-    })
     .epilog('visit https://git.io/vHysA for list of available reporters')
```
mcknasty, line 160: From my understanding, this function should not actually change arguments but only checks to see if they are valid, per the [yargs documentation.](https://yargs.js.org/docs/#api-reference-checkfn-globaltrue)

This condition was never reached per the former code coverage report.

lib/parse-args.js
```diff
@@ -191,18 +188,13 @@ function buildYargs (withCommands = false) {
 function hideInstrumenterArgs (yargv) {
   let argv = process.argv.slice(1)
   argv = argv.slice(argv.indexOf(yargv._[0]))
-  if (argv[0][0] === '-') {
-    argv.unshift(process.execPath)
-  }
   return argv
 }
```
mcknasty, line 192: What use case does this line address?  The only documentation I could find mentioning the `-` is omitting the double dash `--` for configurations that use [package.json or loaded configuration files.](https://github.com/bcoe/c8#cli-options--configuration)   Configuration files for C8 are included in another section of this file.  Does this block of code support loading [configurations via package.json](https://github.com/bcoe/c8#cli-options--configuration)?   Do I need to add a Test Case instead of removing the code? If not, what purpose does this serve?

As the function name suggests this function has something to do with instrumentation arguments.  The only reference I found regarding instrumentation in the documentation was the section [Source map files versus inline source maps.](https://github.com/bcoe/c8#source-map-files-versus-inline-source-maps)

From what I can tell, if the first character of the first argument is a `-` then add the path where C8 is executed to the end of the argument array.

lib/parse-args.js
```diff
@@ -191,18 +188,13 @@ function buildYargs (withCommands = false) {
...
 
 function hideInstrumenteeArgs () {
   let argv = process.argv.slice(2)
   const yargv = parser(argv)
 
-  if (!yargv._.length) return argv
-
   // drop all the arguments after the bin being
   // instrumented by c8.
   argv = argv.slice(0, argv.indexOf(yargv._[0]))
```
mcknasty, line 200: This looks like an early exit.  This condition was never reached per the former code coverage report.  Perhaps another test case?

test/parse-args.js
mcknasty, line 47: Added an additional unit test and changed the title of the describe block.

test/parse-args.js
```diff
@@ -36,6 +36,14 @@ describe('parse-args', () => {
...
```
mcknasty, line 48: Added an additional unit test and changed the title of the describe block.

lib/parse-args.js
```diff
@@ -154,6 +170,10 @@ function buildYargs (withCommands = false) {
     })
     .pkgConf('c8')
     .demandCommand(1)
+    // Todo:  These next 6 lines might need to get removed
+    // This logic is covered with a default argument in
+    // the config options declaration in this file for yargs
+    /* c8 ignore next 6 */
     .check((argv) => {
```
mcknasty, line 177: From my understanding, this function should not actually change arguments but only checks to see if they are valid, per the [yargs documentation.](https://yargs.js.org/docs/#api-reference-checkfn-globaltrue)

This condition was never reached per the former code coverage report.  Added a coverage ignore comment line. 

Setting the tempDirectory when empty should be addressed in this patch on lines 128 -135.

lib/parse-args.js
```diff
@@ -154,6 +170,10 @@ function buildYargs (withCommands = false) {
...
       if (!argv.tempDirectory) {
         argv.tempDirectory = resolve(argv.reportsDir, 'tmp')
```
mcknasty, line 179: From my understanding, this function should not actually change arguments but only checks to see if they are valid, per the [yargs documentation.](https://yargs.js.org/docs/#api-reference-checkfn-globaltrue)

This condition was never reached per the former code coverage report.  Added a coverage ignore comment line. 

Setting the tempDirectory when empty should be addressed in this patch on lines 128 -135.

I suggest these lines can be deleted.

mcknasty, line 179: From my understanding, this function should not actually change arguments but only checks to see if they are valid, per the [yargs documentation.](https://yargs.js.org/docs/#api-reference-checkfn-globaltrue)

This condition was never reached per the former code coverage report.  Added a coverage ignore comment line. 

Setting the tempDirectory when empty should be addressed in this patch on lines 128 -135.

I suggest these lines can be deleted.

lib/parse-args.js
```diff
@@ -191,22 +211,28 @@ function buildYargs (withCommands = false) {
...
-  argv.push(yargv._[0])
+  if (yargv._.length !== 0) {
+    // drop all the arguments after the bin being
+    // instrumented by c8.
+    argv = argv.slice(0, argv.indexOf(yargv._[0]))
+    argv.push(yargv._[0])
+  }
 
   return argv
 }
```
mcknasty, line 238: Change the early exit logic to be more concise and expressive.

lib/parse-args.js
```diff
@@ -5,9 +5,11 @@ const { readFileSync } = require('fs')
 const Yargs = require('yargs/yargs')
 const { applyExtends } = require('yargs/helpers')
 const parser = require('yargs-parser')
-const { resolve } = require('path')
+const { resolve, isAbsolute } = require('path')
 
 function buildYargs (withCommands = false) {
+  let defaultReportDir = resolve(process.cwd(), 'coverage')
```
mcknasty, line 11: Function level variable defaultReportDir that is used in the reports-directory and tmp-directory arguments.

lib/parse-args.js
```diff
@@ -30,7 +32,14 @@ function buildYargs (withCommands = false) {
...
       describe: 'directory where coverage reports will be output to',
-      default: './coverage'
+      default: defaultReportDir,
+      coerce: (opt) => {
+        if (isAbsolute(opt) === false) {
+          defaultReportDir = opt = resolve(process.cwd(), opt)
+        }
+
+        return opt
+      }
```
mcknasty, line 42: These lines ensure absolute paths for the reports directory argument.  It makes use of the [function-level variable](https://github.com/bcoe/c8/pull/447/files?diff=unified&w=0#r1090096554) that could as be required by the temp-directory argument.

lib/parse-args.js
mcknasty, line 179: From my understanding, this function should not actually change arguments but only checks to see if they are valid, per the [yargs documentation.](https://yargs.js.org/docs/#api-reference-checkfn-globaltrue)

This condition was never reached per the former code coverage report.  Added a coverage ignore comment line. 

Setting the tempDirectory when empty should be addressed in this patch on [lines 31 - 42](https://github.com/bcoe/c8/pull/447/files/0eef52ad5780e954315eb37c12235603f498a35c#diff-40d93e4e5f4b94fc8a34a87a3328651c24e855051bfcdd6944dee76f9c82b8d4R31-R42).

I suggest these lines can be deleted.

lib/parse-args.js
bcoe: See the original implementation that this was pulled from:

https://github.com/istanbuljs/nyc/pull/350

I believe it has to do with the `--` operator, which indicates that parsing should stop? There's a test case you can pull over.

mcknasty: Thank you for that pointer @bcoe.  I was able to derive a test case from those examples in that link and those lines were able to get covered.

That code section has to do with undocumented behavior on passing arguments to node.js via C8.  It maybe a backward compatibility issue with previous versions of nyc.

lib/parse-args.js
mcknasty: @bcoe I believe this is the last action item to be resolved before opening up the workflow.  At the moment it has a coverage ignore comment block and removing the block doesn't affect the test results.  I am almost certain that this block of code is covered in [lines 19 - 26](https://github.com/bcoe/c8/pull/447/commits/692ec75f628fddc426ddda6a09f2a7ca95240611#diff-40d93e4e5f4b94fc8a34a87a3328651c24e855051bfcdd6944dee76f9c82b8d4R19-R26) and [line 149](https://github.com/bcoe/c8/pull/447/commits/692ec75f628fddc426ddda6a09f2a7ca95240611#diff-40d93e4e5f4b94fc8a34a87a3328651c24e855051bfcdd6944dee76f9c82b8d4R149).  Thoughts?

test/parse-args.js
```diff
@@ -44,14 +64,16 @@ describe('parse-args', () => {
       argv.lines.should.be.equal(95)
     })
     it('should use config file specified in --config', () => {
+      const tmpDir = resolve(process.cwd(), 'foo')
       const argv = buildYargs().parse(['node', 'c8', '--config', require.resolve('./fixtures/config/.c8rc.json')])
       argv.lines.should.be.equal(101)
-      argv.tempDirectory.should.be.equal('./foo')
+      argv.tempDirectory.should.be.equal(tmpDir)
```
bcoe, line 70: I don't like that we have a behavioural change as a result of these updates, coverage should ideally just be changing coverage.

I don't remember if there's a good reason why I don't `resolve` the path of `tmpDir`, but if there are no outstanding bugs resulting from the relative path, I don't think we should change it (especially in a PR that is simply attempting to increase coverage).

lib/parse-args.js
```diff
@@ -154,6 +174,10 @@ function buildYargs (withCommands = false) {
     })
     .pkgConf('c8')
     .demandCommand(1)
+    // Todo:  These next 6 lines might need to get removed
+    // This logic is covered with a default argument in
+    // the config options declaration in this file for yargs
+    /* c8 ignore next 6 */
     .check((argv) => {
```
bcoe, line 181: I don't understand the reason for introducing the new coerce logic, vs., just adding a test to cover the `check` behaviour.

I would rather we don't introduce coerce behaviour to the arguments.

bcoe (changes requested):
Left a couple nits, I'd rather avoid changing behaviour around path resolution. The reason being that people have complicated setups with c8, where they run it in different directories, run it against output from web browsers (which may have files within them that are not relative to the director that c8 is running in, etc).

tldr; it might be okay to `resolve` the reportsDir/tmpDir, but it may have unintended consequences upstream.

Perhaps I'm being too cautious, but I'd rather not change this part of the library (unless we're trying to address outstanding bugs, and have a regression test).

lib/parse-args.js
bcoe: > From my understanding, this function should not actually change arguments but only checks to see if they are valid

It might be less confusing to `corece` the temp directory, but we should only run the `resolve` logic if temp directory isn't set, to match the existing logic. And we shouldn't do any coercion on the report directory.

mcknasty:
The issue of the check function mentioned above has come up a few times as I have been working on yargs related feature requests and code coverage reports.  I outlined the issue as I saw it in #489.  Furthermore, merging pull request #436 could change the delta of this pull request a fair amount.  I suggest that I outline the various ways of setting the temp directory and then write unit tests against those scenarios.   I think we can revisit this issue once these few dependencies are complete.

test/parse-args.js
```diff
@@ -34,8 +46,20 @@ describe('parse-args', () => {
...
+    it('should set it if undefined', () => {
+      const NODE_V8_COVERAGE = process.env.NODE_V8_COVERAGE
+      process.env.NODE_V8_COVERAGE = undefined
+      process.argv = ['node', 'c8', '--foo=99', 'my-app', '--help']
+      const argv = buildYargs().parse(hideInstrumenteeArgs())
+      argv.tempDirectory.endsWith('/coverage/tmp').should.be.equal(true)
+      delete process.env.NODE_V8_COVERAGE
       process.env.NODE_V8_COVERAGE = NODE_V8_COVERAGE
     })
+    /** */
```
mcknasty, line 62: Another candidate for updating unit tests as it relates to (#516)

mcknasty, line 62: Need to uncomment and skip this test.

PR #517, opened 2024-01-20
feat: show derived configuration

mcknasty:
# **[Pull Request 517](https://github.com/bcoe/c8/pull/517)**

## Commit Message

commit b33e8e7773ea79897c8b77f92894a7cb3fc5aaf7
Author: Trevor D McKeown <tmckeown@gmail.com>
Date:   Sun Feb 18 09:17:58 2024 -0500

    Print Derived Configuration Report (#517)

    feat: print derived config variables
    feat: print derived config variables in json
    test: 16 new unit tests to support features
    test: 1 skipped unit test for discovered bug in yargs with reports param

## **Following the Contributions Recommendations [here.](https://github.com/bcoe/c8/blob/main/CONTRIBUTING.md)**

1. [x] Make sure you have installed the latest version of Node.js
2. [x] Fork this project on GitHub and clone your fork locally
3. [x] Create local branches to work within. These should also be created directly from the main branch.  Local Fork [here.](https://github.com/mcknasty/c8)
4. [x] Make your changes.
5. [x] Run tests to make sure all is okay (everything should pass except the snapshot).
   1. A [complete log](https://gist.github.com/mcknasty/465e70521191d4ab96906d70fccf0b0d#file-initial-test-txt) of initial test results.
   2. As instructed, ignore snapshot failures. [1 pending](https://gist.github.com/mcknasty/465e70521191d4ab96906d70fccf0b0d#file-initial-test-txt-L172)
   3. [111 passing](https://gist.github.com/mcknasty/465e70521191d4ab96906d70fccf0b0d#file-initial-test-txt-L171) in [54 seconds](https://gist.github.com/mcknasty/465e70521191d4ab96906d70fccf0b0d#file-initial-test-txt-L171)
6. [x] Now update the snapshot.
   1. A [complete log](https://gist.github.com/mcknasty/465e70521191d4ab96906d70fccf0b0d#file-snapshot-test-txt) of snapshot test results.
   2. [111 passing](https://gist.github.com/mcknasty/465e70521191d4ab96906d70fccf0b0d#file-snapshot-test-txt-L175) in [54 seconds](https://gist.github.com/mcknasty/465e70521191d4ab96906d70fccf0b0d#file-snapshot-test-txt-L175)
7. [x] If all is passing, commit your changes. The log of the commit can be found [here.](https://github.com/bcoe/c8/pull/517/commits/32dd6afec6a19eb36c71a3189991775792207057)
[...]

Issue #425, opened 2022-11-04
Allow configuration file to be extended


blephy:
I,

It would be great to be able to `extends` a configuration file to another, like `TSConfig` or `Eslint` do, or at least to be able to write `.cjs` / `.js` configuration file to require another `js` file.

We use c8 in a mono-repo, and we cannot share a c8 configuration file between workspace:/

Thanks

bcoe:
This seems like a good feature request 👍

blephy:
Do you need some help ?

blephy:
@bcoe i did look to the c8 and yargs code. In fact, we can already extends configuration if the path to the extendable config is a `json` file.

This is already working : 

```json
{
  extends: '../../config/c8/c8.config.json',
  // ...
```

What about making c8 compatible with yml and js configuration ? [cosmiconfig](https://github.com/davidtheclark/cosmiconfig) sounds good for this.

bcoe:
Closing in favour of:

https://github.com/bcoe/c8/issues/431

With regards to cosmicconfig, I would prefer not to pull in a dependency for config management, if it can be avoided. I think it's not worth the complexity it would add to the library.

> Additionally, we should update the documentation. I did a branch but cannot push to this repository :/

You will need to create a pull request form a fork, [here's some documentation on how to do so](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request-from-a-fork) 👍

Issue #420, opened 2022-10-10
Feature request: Add support for configuration via something that allows comments


kf6kjg:
[JSON](https://www.json.org/json-en.html) specifically doesn't allow for comments, since it's designed for data transfer not for configuration, and c8 crashes if the config file has any. This is currently expected, but isn't preferred. 

Possible alternatives:
- JSONC: Just JSON, but adding syntax support for C/C++ style comments. Probably the simplest fix as it merely involves either changing out [`JSON.parse`](https://github.com/bcoe/c8/blob/e81ed5dd9ef5dac1a1f2f6dc26a07abb6c05d709/lib/parse-args.js#L18) for a library that supports comments in JSON or by DIY'ing it by pre-filtering the file contents,. though I DO NOT recommend that latter approach. This might need to be handled via also allowing the file to have the extension `jsonc` so that editors figure it out too.
- YAML: used extensively by many projects, including nyc. Grammatically similar enough to JSON that there's almost no impedance mismatch.

Consider the following:

```JSON
{
    "branches": 70,
    "functions": 75,
    "lines": 80,
    "statements": 80
}
```

This tells us nothing of the goals of the project or why things are the way they are vs the following, nor does it allow for easy in-code tracking of what needs to be changed:

```JSONC
{
    // 80% minimums per company standards.
    "branches": 70 /* TODO: Increase to at least 80 ASAP, was set to match status quo when coverage checking was added to the code. */,
    "functions": 75 /* TODO: Increase to at least 80 ASAP, see above for why. */,
    "lines": 80,
    "statements": 80
}
```

Yes, you can argue that issues should be tracked in an issue tracker and that historical notes could be handled by reviewing the comments in the git history. I agree, however there's nothing like having the problem noted and in your face every time you look at the relevant block of code. Or if you are on a jet disconnected from the internet and want to fix some low hanging fruit.

bcoe:
Closing in favour of https://github.com/bcoe/c8/issues/431, thanks for the feature request!

PR #518, opened 2024-02-02
Additional Configuration File Support (#436)

mcknasty:
#Temp PR Message

feat: Adding additional configuration file options
test: unit test for config file detection
test: unit test for config files pass via the cli
test: unit test case for early exit of hideInstrumenteeArgs
test: unit test for Node.js argument handling
fix: bug in spawning of c8 process that dropped coverage significantly

##### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,12 +18,24 @@ The above example will output coverage metrics for `foo.js`.
 
 ## CLI Options / Configuration
 
-c8 can be configured via command-line flags, a `c8` section in `package.json`, or a JSON configuration file on disk.
-
-A configuration file can be specified by passing its path on the command line with `--config` or `-c`. If no config option is provided, c8 searches for files named `.c8rc`, `.c8rc.json`, `.nycrc`, or `.nycrc.json`, starting from
-`cwd` and walking up the filesystem tree.
+c8 can be configured via command-line flags, a `c8` section in `package.json`, or a configuration file on disk.
 
 When using `package.json` configuration or a dedicated configuration file, omit the `--` prefix from the long-form of the desired command-line option.
+
+A configuration file can be specified by passing its path on the command line with `--config` or `-c`. If no config option is provided, c8 searches for files named in the table below starting from `cwd` and walking up the filesystem tree.
+
+A robust configuration file naming convention is available in an effort to stay compatible with nyc configuration options and ensure dynamic configuration.
+
+| File name                                                                                           | File Association   |
+|-----------------------------------------------------------------------------------------------------|--------------------|
+| `.c8rc`, `.c8rc.json`                                                                               | JSON               |
+| `.c8rc.yml`, `.c8rc.yaml`                                                                           | YAML               |
+| `.c8rc.js`, `.c8rc.cjs`, `.c8.config.js`, `.c8.config.cjs`, `c8.config.js`, `c8.config.cjs`         | CommonJS export*   |
+| `.nycrc`, `.nycrc.json`                                                                             | JSON               |
+| `.nycrc.yaml`, `.nycrc.yml`                                                                         | YAML               |
+| `.nycrc.js`, `.nycrc.cjs`, `nyc.config.js`, `nyc.config.cjs`, `.nyc.config.js`, `.nyc.config.cjs`   | CommonJS export*   |
+
+For packages written in ESM module syntax, a static configuration option is supported in JSON or YAML syntax. A dynamic configuration is also supported.  These configuration files must be written in CommonJS utilizing one of the .cjs file options in the table above. At the moment ESM syntax is not supported for writing c8 configuration files.  This may change in the future, but please note, C8 is written in CommonJS syntax.  
 
 Here is a list of common options. Run `c8 --help` for the full list and documentation.
 
--- a/lib/parse-args.js
+++ b/lib/parse-args.js
@@ -5,7 +5,7 @@ const { readFileSync } = require('fs')
 const Yargs = require('yargs/yargs')
 const { applyExtends } = require('yargs/helpers')
 const parser = require('yargs-parser')
-const { resolve } = require('path')
+const { resolve, extname, basename } = require('path')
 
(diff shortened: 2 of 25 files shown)


mcknasty:
Reference to old PR (#436)

bcoe:
I'm hesitant to iuntroduce `js-yaml` dependency which is fairly large and doesn't seem to be updated frequently. Also, historically yaml is a complex enough DSL that it is known to have serious security vulnerabilities in some implementations.

A couple options:

- Could we allow users to bring their own copy of `js-yaml` and print a warning if they try to use `yaml` configuration without installing the dependency (`to use yaml configuration please install js-yaml as a dev dependency`) do any other tools do something like this?
- Long term solution, a yaml parser seems like something worth considering in Node core.

kf6kjg:
By moving JS-yaml to being lazy loaded we could make it an optional peer dependency. Then if the user needs it they can install it. Everyone else doesn't get it in their tree.

bcoe:
(_@mcknasty let's keep this PR open and drive it to completion, thank you for your contributions_)

@kf6kjg @mcknasty, I'm very much on the fence with yaml. One one hand, I understand users would love to be able to use yaml. On the other hand, I spend most of my OSS time updating dependencies to deal with transitive security issues, and the fact that the `yaml` libraries are a fairly large surface makes me consider them a good potential candidate to get CVEs logged against them.

Lazy loading is perhaps a good approach, do you think users would be willing to add `js-yaml` as a dev dependency if we throw an error that explains how?

kf6kjg:
IIRC those were the goals.

mcknasty:
I would like to request some clarity on the comments above. We don't need to explicitly [include js-yaml](https://gist.github.com/mcknasty/1d666798cf45232e115439203ddedf97#file-c8-js-yaml-dep-bash-example-txt-L27-L39) as a development dependency.  The unit tests [pass](https://gist.github.com/mcknasty/1d666798cf45232e115439203ddedf97#file-c8-js-yaml-dep-bash-example-txt-L63-L80) for this [commit](https://gist.github.com/mcknasty/1d666798cf45232e115439203ddedf97#file-c8-js-yaml-dep-bash-example-txt-L24) , if I remove `js-yaml` from `package.json`.  [The package is still included](https://gist.github.com/mcknasty/1d666798cf45232e115439203ddedf97#file-c8-js-yaml-dep-bash-example-txt-L52-L60) in the `node_module` directory due to the fact it's a dependency of [eslint](https://gist.github.com/mcknasty/1d666798cf45232e115439203ddedf97#file-c8-js-yaml-dep-bash-example-txt-L55-L60).  Furthermore, [eslint is a dependency](https://github.com/standard/standard/blob/16af4a2811ccf27980e4b5ad2d68a89918bd09fc/package.json#L18) of standard.

@bencoe Are you planning on moving away from standard as a lint tester?  Is that why we want to prompt for an optional download or path for a yaml parsing binary?

Additionally, this pull request has a dependency on #517.  The print derived config option is required to verify what type of file has been loaded by yargs in the [pending unit tests](https://gist.github.com/mcknasty/1d666798cf45232e115439203ddedf97#file-c8-js-yaml-dep-bash-example-txt-L63-L80).

@bcoe, #517 was at [100% coverage](https://gist.github.com/mcknasty/9256f73327975964e5a6cdd84779b0c9#file-print-mess-config-pr-txt-L33) and I am fairly certain it was passing through the CD/CI.  There were only two concerns we were working through.  Would you like me to open a new pull request?  There was [a comment](https://github.com/bcoe/c8/pull/517#discussion_r1495014684) about test setup functions.  Is there code style guide you could refer me to?

If there is anything additional you would like to discuss, please feel free to email me at the email address in the commit log.

kf6kjg:
IMO being a transitive dependency is fine. Ifnwhen those packages decide to remove js-yaml, or even make it an optional peer dependency like I suggested, then it could disappear entirely. In the mean time this package only declares that if you need it you can install and use it. In my case I probably wont, I want the CJS configs these days! :D

mcknasty:
`js-yaml` doesn't require `c8`.  Is there a reason it needs to be a peer dependency?  Can it just be an [optional dependency](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#optionaldependencies)?

kf6kjg:
Thank you for pointing out optional dependencies, that one I'd not directly worked with before so it was good to learn.

From what I learnt: Optional dependencies mean that if it is omitted via the `--omit=optional` NPM flag it will not be installed, but if it is then if there is a conflicting versions specified by the root package or some other package then `c8` will get its own copy.

Additionally peer dependencies are more [for "plugin" architectures](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#peerdependencies), which in this case `c8` isn't doing: we aren't passing to `c8` an instance of `js-yaml`. Thus I think that you are correct in pointing out that being an optional dependency is the more correction option.

kf6kjg:
I love TOML, used it myself one a long time back, but it's never gotten the traction that YAML has and it would be a lot harder on the users to figure out. That said, while TOML solves #420, I've come to the conclusion that I'm more a fan of #425's JS-based, and [TS-based in the future](https://tsx.is/node/ts-import), configs which as a side effect ALSO solves the basic want of #420 - comments. Thus if YAML is full blocker for you, though I'm still not sure why, and @mcknasty doesn't feel like implementing TOML then I'd say we cull the YAML support and get the ball rolling on the JS support.

bcoe:
>  Thus if YAML is full blocker for you, though I'm still not sure why, and @mcknasty doesn't feel like implementing TOML then I'd say we cull the YAML support and get the ball rolling on the JS support.

This sounds like a great compromise to me, I like that `.js` and `.mjs` support wouldn't require us pulling in a beefy library. And it should address many of the same requirements.

Sorry for being such a pain in the neck on this PR. I've grown over the years to be much more conservative with dependency trees.

kf6kjg:
@mcknasty it's been a while. You still open to seeing this PR through? I'm back working on things that would benefit from the JS-based config.

Issue #502, opened 2023-11-22
Coverage information too high for large esbuild produced bundle


jakebailey:
* **Version**: v20.9.0
* **Platform**: Linux JABAILE-DESK02 5.15.90.1-microsoft-standard-WSL2 #1 SMP Fri Jan 27 02:56:13 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

I can only apologize in advance for this one...

I work on [TypeScript](http://github.com/microsoft/TypeScript), and I've been wanting to set up coverage information for the compiler. c8 does a great job; significantly faster than nyc (of course). I use c8 in my own projects with success.

However, capturing coverage for the compiler produces results that are too high to believe. `checker.ts` is a 50k line file, and yet its coverage is:

```
checker.ts | 99.99 | 95.38 | 98.62 | 99.99 | 50577-50578
```

There's no way that we actually have tested every line of the checker. And indeed, adding a function to a known-dead path (like one ending in `Debug.assertNever`) indicates that the line is covered when it definitely is not, and the overall count doesn't change.

![image](https://github.com/bcoe/c8/assets/5341706/9183aa93-4c10-49ad-a326-b80c5c804088)

I don't know if I have a good repro for this besides `npx c8 npm test -- --no-lint` on our repo. Obviously that's huge but I don't really know where to start.

At best, I can observe that `npx c8 npm test -- --no-lint --no-bundle` works better, i.e. not using esbuild and instead running the tsc output produces a more believable coverage count:

```
checker.t  |    97.5 |    95.87 |   98.62 |    97.5 | ...-48684,48686-48687,48761-48762,48 ...
```

jakebailey:
One thing that is interesting is that comparing the bundled output HTML page with the unbundled output HTML page, the paths are different. Specifically, the bundled output starts right with `compiler/checker.ts`, whereas the unbundled output has `src/compiler/checker.ts`.

jakebailey:
Yeah, the unbundled output is way better:

![image](https://github.com/bcoe/c8/assets/5341706/8fb25c12-e1f8-415e-ad41-98e5e0e58512)

ericmorand:
@jakebailey did you happen to solve the issue? I have the same problem with bundled test suite where, as soon as a portion of a source file is covered, the whole file is marked as covered.

jakebailey:
No, besides using https://github.com/bcoe/c8/pull/521.

PR #521, opened 2024-02-20
feat: support monocart report

cenfun:
@bcoe This PR is to add [Monocart](https://github.com/cenfun/monocart-coverage-reports) report support for c8.

In order to adapt to c8, Monocart has made many adjustments, such as supporting Node 14 and allowing empty coverage for `--all` option, and so on.
Finally, it can pass the test today.
Please let me know if there are any issues, and I will provide more information as necessary.
Thanks

##### Checklist


bcoe:
@cenfun thanks for the contribution, this looks really good.

Question, how do you represent the empty coverage in Istanbul format? This seems like something I could fix in `v8-to-istanbul` as well?

Currently this is what an uncovered file looks like when running `--all`:

```json
{
  "/Users/bencoe/google/oss/c8-515/foo.js": {
    "path": "/Users/bencoe/google/oss/c8-515/foo.js",
    "all": true,
    "statementMap": {
      "0": {
        "start": {
          "line": 1,
          "column": 0
        },
        "end": {
          "line": 1,
          "column": 0
        }
      }
    },
    "s": {
      "0": 1
    },
    "branchMap": {},
    "b": {},
    "fnMap": {},
    "f": {}
  }
```

jakebailey:
This is a bit of a drive-by, but I tried this PR out in hopes it would fix #502; I get loads of these messages:

```
[MCR] failed to parse source file: src/services/formatting/formatting.ts Unexpected token (82:7)
[MCR] failed to parse source file: src/services/formatting/formattingContext.ts The keyword 'enum' is reserved (14:13)
[MCR] failed to parse source file: src/services/formatting/formattingScanner.ts Unexpected token (29:7)
[MCR] failed to parse source file: src/services/formatting/rule.ts Unexpected token (10:7)
[MCR] failed to parse source file: src/services/formatting/rules.ts Unexpected token (37:7)
[MCR] failed to parse source file: src/services/formatting/rulesMap.ts Unexpected token (19:40)
[MCR] failed to parse source file: src/services/formatting/smartIndenter.ts Unexpected token (62:7)
[MCR] failed to parse source file: src/services/getEditsForFileRename.ts Unexpected token (55:11)
[MCR] failed to parse source file: src/services/goToDefinition.ts Unexpected token (115:47)
```

Which seems problematic (given how much code out there is actually TypeScript).

But, the coverage looks more reasonable! With the experimental flag:

```
  checker.ts                                              |   95.21 |    95.41 |    98.4 |   95.29 | ...49050,49100,49114,49163,49199,49220-49223,49298,49323-49326,49342,49374,49540,49545,49706,49940,50204,50375,50551,50602,50926,50947-50948,50960,51069,51079,51116,51146,51153,51160,51195,51202 
```

Compare to c8 as-is (definitely wrong):

```
  checker.ts                                              |   99.99 |    95.38 |   98.68 |   99.99 | 51117-51118        
```

And in our `--no-bundle` mode (which runs tests without esbuild-ing it, so is very slow at runtime):

```
  checker.ts                                              |   97.56 |    95.88 |   98.68 |   97.56 | ...49221-49222,49224-49225,49299-49300,49324-49325,49327-49328,49343-49344,49375-49376,49546,49707-49708,49941,50004-50005,50205-50206,50603-50604,50927-50929,50948-50952,50961-50962,51117-51118 
```

Not totally sure what the 2% difference is, but I'm happy that maybe there's a shred of hope to get good coverage info without nyc.

jakebailey:
Well, the output coverage is a little... weird: 

image

cenfun:
@bcoe Yes, agree. Checking the coverage of untested files is truly a challenge, especially for checking `functions` and `branches`. For example `--branches=80` it requires to know:
- 1, the total of branches. 
- 2, the percent of branches (uncovered branches / total)

And the total of branches might be 0 if a file doesn't have any branches. At this point, the percentage doesn't make sense, we should skip the checking when total = 0, or it can be simply assumed that the percent is 100%.

The current challenge is that we don't know the total count of these metrics for the untested files.
- For `v8-to-istanbul`, it doesn't parse the files, hence it doesn't know the total.
- For Monocart, it can parse JavaScript into AST, so it can get the total. However, the untested files could be in any format, such as `.ts`, `.vue`, `.svelte` and so on, so it cannot be successfully parsed if it's not in the standard ECMA format. (this is why @jakebailey got the parser error message, but I have removed this message in version [2.5.4](https://www.npmjs.com/package/monocart-coverage-reports/v/2.5.4))

Anyway, I made a small fixing, see [commit](https://github.com/bcoe/c8/pull/521/commits/b6a086294c12bff2da0a8499ded6b260f540c128)
```js
if (file && file.empty) {
    process.exitCode = 1
    console.error(
      'ERROR: Empty coverage (untested file) does not meet threshold for ' +
        path.relative('./', file.sourcePath).replace(/\\/g, '/')
    )
    return
  }
```
When the checked file is an empty coverage, it always prompts that the threshold is not passed. Is this okay?

@jakebailey Thanks for the feedback. Can you provide a small repo which reproduces your issue?  Perhaps I can help fix it.

jakebailey:
> Can you provide a small repo which reproduces your issue? Perhaps I can help fix it.

I'm afraid I don't have a small repo; I'm running this on TypeScript itself: https://github.com/microsoft/TypeScript

Using `c8 npm test -- --no-lint`.



> I noticed that in your screenshot you are using the Istanbul's html report, which requires converting v8 format to Istanbul, so there might be some other issues such as problems with sourcemaps. Could you try the native `v8` report?

I tried with those flags, and get a similar result in the console:

```
checker.ts │  95.24 % │    95.22 % │  95.42 % │   98.41 % │  93.39 % │ 1511-1519,1530,1535,1542-1550,1563,1586,1595,16...
```

The HTML output looks a lot better in this PR than with monocart-coverage-reports alone (which did not correctly handle source maps):

image

Though it's not super clear to me in general how to read this info compared to strictly "hits per line".

But, I'm personally quite happy with this PR.

Scrolling through, there are some oddities, like:

image

image

Running `c8 --experimental-monocart --reporter=v8 --reporter=console-details npm test -- --no-lint` will get you a result, if you're looking for a stress test.

cenfun:
@jakebailey thank you for providing the test case. I was able to fix few more issues in version [2.5.5](https://www.npmjs.com/package/monocart-coverage-reports/v/2.5.5)
- fixed UI performance issue. There is a file `checker.ts` is so big (2.8M), it could cause the rendering to freeze.
- fixed the sourcemap issue. The sourcemap file `run.js.map` does not include the `sourcesContent`. For c8, the `cource-map-cache` does not include the source code but `lineLengths` (fake source).
- some UI minor changes. 
For the green `x number` button, it is actually the execution counts on its range, mouse hover on the green button will show the range. Just add a `count` switch button to display or hide the execution counts if it's too much and affects readability. 
For `hits per line`, I thought that the native V8 coverage provides coverage for bytes, not lines, perhaps it's still not quite accustomed.
image

jakebailey:
Thanks! I didn't test with `mcr` directly, just this PR again, and it all seems to be working from what I can tell. Looking forward to having something that works.

ericmorand:
@cenfun I published a minimal repository to test c8 with bundles - namely rollup generated bundles: https://github.com/ericmorand/c8-with-bundle

This PR seems to fix C8 issues but I'll perform some more testings with some edge cases.

bcoe:
@ericmorand how has your testing been going. This is a pretty big addition to the library, which is one reason I've been hesitant to land. But, if it's solving user problems, I think we should figure out how to expose Monocart.

@cenfun my other hesitancy with landing has been that `lib/monocart-report.js` seems to essentially introduce an alternate path through the codebase that repeats logic similar to `report.js` (my hope was that we could instead figure out some minor changes to the existing contract in `report.js` that would perhaps work with both Monocart or v8-to-istanbul). My concern is maintenance burden, since both files do similar things like instantiating `test-exclude`, rewriting file paths

cenfun:
@bcoe good suggestion, thanks. I've already updated it, please review it.

README.md
```diff
@@ -119,6 +120,12 @@ The `--100` flag can be set for the `check-coverage` as well:
 c8 check-coverage --100
 ```
 
+## Using Monocart coverage reports (experimental)
+[Monocart](https://github.com/cenfun/monocart-coverage-reports) will bring additional support for native V8 coverage reports, for example:
```
bcoe, line 124: I think we should explain what monocart is a bit better. It's an alternate tool for converting raw V8 coverage reports to other coverage formats, as an alternate to `v8-to-istanbul` and istanbul reports.

lib/report.js
```diff
@@ -89,6 +95,128 @@ class Report {
     }
   }
 
+  async runMonocart () {
```
bcoe, line 98: 👍 I prefer now that we're sharing some of the same behaviour in `report.js`.

README.md
cenfun: I have added some explanations, please take a look.

README.md
```diff
@@ -119,6 +120,17 @@ The `--100` flag can be set for the `check-coverage` as well:
...
+## Using Monocart coverage reports (experimental)
+Monocart will bring additional support for native V8 coverage reports, for example:
+```sh
+c8 --experimental-monocart --reporter=v8 --reporter=console-details node foo.js
+```
+There are some differences as follows:
+- `c8` uses `v8-to-istanbul` by default to convert V8 coverage data and generate Istanbul reports.
+- `c8 --experimental-monocart` will use Monocart's new converter to convert V8 coverage data and generate Istanbul reports, it could provide better data for reports. Moreover, it can also directly generate V8 coverage reports (based on Bytes), such as available reports: `v8`, `console-details`, etc.
+
+See [monocart-coverage-reports](https://github.com/cenfun/monocart-coverage-reports) for more info.
```
bcoe, line 132:
```suggestion
Monocart is an alternate library for outputting [v8 code coverage](https://v8.dev/blog/javascript-code-coverage) data as Istanbul reports.

Monocart also provides reporters based directly on v8's byte-offset-based output. Such as, `console-details` and `v8`. This removes a complex transformation step and may be less bug prone for some environments.

**Example usage:**

```sh
c8 --experimental-monocart --reporter=v8 --reporter=console-details node foo.js
```

See [monocart-coverage-reports](https://github.com/cenfun/monocart-coverage-reports) for more info.
```

Comment (cenfun) [reply README.md]: Thanks a lot

**Comment by bcoe** [comment]
@cenfun left a suggestion for wording. could I also bother you to rebase?

**Comment by cenfun** [comment]
@bcoe I'm not very good at writing documents, please feel free to modify it,  thank you.

File: package.json [hunk 4]
```diff
@@ -40,6 +40,7 @@
     "istanbul-lib-coverage": "^3.2.0",
     "istanbul-lib-report": "^3.0.1",
     "istanbul-reports": "^3.1.6",
+    "monocart-coverage-reports": "^2.8.3",
```
bcoe, line 43: similar to what is being discussed [here](https://github.com/bcoe/c8/pull/518) let's make monocart a `peerDependency` and update our docs to include:

```
npm i monocart-coverage-reports --save-dev
```

I think this is the right call since monocart is fairly large:

```
npm notice package size:  253.8 kB                                
npm notice unpacked size: 905.1 kB   
```

And I don't think it adds that much of a hinderance to people wanting to use it.

bcoe: For reference: https://flaviocopes.com/npm-peer-dependencies/

bcoe:
> @bcoe I'm not very good at writing documents, please feel free to modify it, thank you.

@cenfun I thought your docs for for Monocart are very good. I like how you approached having READMEs in more than one language, with a link at the top.

package.json
cenfun: I moved it to peerDependencies, it good a suggestion. Due to the new V8's report having a very complex UI, it take a large size.

lib/parse-args.js
```diff
@@ -158,6 +158,11 @@ function buildYargs (withCommands = false) {
       describe: 'supplying --merge-async will merge all v8 coverage reports asynchronously and incrementally. ' +
         'This is to avoid OOM issues with Node.js runtime.'
     })
+    .option('experimental-monocart', {
+      default: false,
+      type: 'boolean',
+      describe: 'Using Monocart coverage reports'
```
bcoe, line 164:
```suggestion
      describe: 'Use Monocart coverage reports'
```

lib/report.js
```diff
@@ -16,6 +16,7 @@ const getSourceMapFromFile = require('./source-map-from-file')
 const v8toIstanbul = require('v8-to-istanbul')
 const util = require('util')
 const debuglog = util.debuglog('c8')
+const { CoverageReport } = require('monocart-coverage-reports')
```
bcoe, line 19: we should move this into a try/catch.

lib/report.js
```diff
@@ -73,6 +76,9 @@ class Report {
   }
 
   async run () {
+    if (this.monocartArgv) {
+      return this.runMonocart()
```
bcoe, line 80: If we were unable to require Monocart, we should print the install instructions here and exit with 1.

lib/report.js
bcoe: Perhaps we should move if into a try/catch in a separate monocart.js helper file, then we can mock that file to test the help output.

bcoe:
@cenfun I left some more review. If you could also resolve conflicts, I think this is getting close to ready.

lib/report.js
cenfun: I have made the following improvements based on your suggestion
- Move the import library to try/catch.
- Use the `import("xxx")` function to load the lib asynchronously. It should be good enough without creating a new help file, and added new test case to cover it (rename the lib entry file can mock it).
- Print the install instructions and exit with 1 when failed to load the lib.

Please review, thanks.

lib/report.js
```diff
@@ -89,6 +94,137 @@ class Report {
     }
   }
 
+  async getMonocart () {
+    let MCR
+    try {
+      MCR = await import('monocart-coverage-reports')
+    } catch (e) {
+      console.error('ERROR: Failed to import module "monocart-coverage-reports". Please try to install it: npm i monocart-coverage-reports --save-dev')
```
bcoe, line 102:
```suggestion
      console.error('--experimental-monocart requires the plugin monocart-coverage-reports. Run: "npm i monocart-coverage-reports --save-dev"')
```

bcoe:
@cenfun running the test suite once deletes the `index.js` file in monocart reports:

```
bash-3.2$ cat node_modules/monocart-coverage-reports/lib/index.
index.d.ts    index.js.bak  index.mjs  
```

I think we should drop this logic from tests:

```js
        const mcrPath = require.resolve('monocart-coverage-reports')
        rename(mcrPath, mcrPath + '.bak', (err) => {
          if (err) { throw err }
        })
```

It seems fragile and I think you could just mock a method like this:

```js
  async getMonocart () {
    let MCR
    try {
      MCR = await this.importMonocart();
    } catch (e) {
      console.info(e);
      console.error('--experimental-monocart requires the plugin monocart-coverage-reports. Run: "npm i monocart-coverage-reports --save-dev"')
      process.exit(1)
    }
    return MCR
  }

  async importMonocart () {
    await import('monocart-coverage-reports')
  }
```

so, mock `importMonocart` to throw.

cenfun:
@bcoe it‘s done, please review.

bcoe:
@cenfun this looks good to me, but I'm having test failures locally:

```
2 failing

  1) c8
       --extension
         includes coverage when extensions specified with --all:

      AssertionError: expected value to match snapshot c8 --extension includes coverage when extensions specified with --all 1
      + expected - actual

       hey
       ---------------------------------------|---------|----------|---------|---------|------------------------
       File                                   | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s      
       ---------------------------------------|---------|----------|---------|---------|------------------------
      -All files                              |     3.2 |    11.11 |    5.76 |     3.2 |                        
      +All files                              |    3.21 |    12.24 |    6.38 |    3.21 |                        
        c8                                    |       0 |        0 |       0 |       0 |                        
         index.js                             |       0 |        0 |       0 |       0 | 1                      
        c8/bin                                |       0 |        0 |       0 |       0 |                        
         c8.js                                |       0 |        0 |       0 |       0 | 1-43                   
        c8/coverage                           |       0 |        0 |       0 |       0 |                        
         block-navigation.js                  |       0 |        0 |       0 |       0 | 1-87                   
      -  coverage-data.js                     |       0 |        0 |       0 |       0 | 1                      
         prettify.js                          |       0 |        0 |       0 |       0 | 1-2                    
         sorter.js                            |       0 |        0 |       0 |       0 | 1-196                  
      - c8/coverage/assets                    |       0 |        0 |       0 |       0 |                        
      -  monocart-code-viewer.js              |       0 |        0 |       0 |       0 | 1                      
      -  monocart-coverage-app.js             |       0 |        0 |       0 |       0 | 1                      
      -  monocart-formatter.js                |       0 |        0 |       0 |       0 | 1                      
      -  turbogrid.js                         |       0 |        0 |       0 |       0 | 1                      
        c8/lib                                |       0 |        0 |       0 |       0 |                        
         parse-args.js                        |       0 |        0 |       0 |       0 | 1-229                  
         report.js                            |       0 |        0 |       0 |       0 | 1-542                  
         source-map-from-file.js              |       0 |        0 |       0 |       0 | 1-100                  
      
      at Context.<anonymous> (test/integration.js:697:40)
      at process.processImmediate (node:internal/timers:476:21)

  2) c8 mergeAsync
       --extension
         includes coverage when extensions specified with --all:

      AssertionError: expected value to match snapshot c8 mergeAsync --extension includes coverage when extensions specified with --all 1
      + expected - actual

       hey
       ---------------------------------------|---------|----------|---------|---------|------------------------
       File                                   | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s      
       ---------------------------------------|---------|----------|---------|---------|------------------------
      -All files                              |     3.2 |    11.11 |    5.76 |     3.2 |                        
      +All files                              |    3.21 |    12.24 |    6.38 |    3.21 |                        
        c8                                    |       0 |        0 |       0 |       0 |                        
         index.js                             |       0 |        0 |       0 |       0 | 1                      
        c8/bin                                |       0 |        0 |       0 |       0 |                        
         c8.js                                |       0 |        0 |       0 |       0 | 1-43                   
        c8/coverage                           |       0 |        0 |       0 |       0 |                        
         block-navigation.js                  |       0 |        0 |       0 |       0 | 1-87                   
      -  coverage-data.js                     |       0 |        0 |       0 |       0 | 1                      
         prettify.js                          |       0 |        0 |       0 |       0 | 1-2                    
         sorter.js                            |       0 |        0 |       0 |       0 | 1-196                  
      - c8/coverage/assets                    |       0 |        0 |       0 |       0 |                        
      -  monocart-code-viewer.js              |       0 |        0 |       0 |       0 | 1                      
      -  monocart-coverage-app.js             |       0 |        0 |       0 |       0 | 1                      
      -  monocart-formatter.js                |       0 |        0 |       0 |       0 | 1                      
      -  turbogrid.js                         |       0 |        0 |       0 |       0 | 1                      
        c8/lib                                |       0 |        0 |       0 |       0 |                        
         parse-args.js                        |       0 |        0 |       0 |       0 | 1-229                  
         report.js                            |       0 |        0 |       0 |       0 | 1-542                  
         source-map-from-file.js              |       0 |        0 |       0 |       0 | 1-100                  
      
      at Context.<anonymous> (test/integration.js:697:40)
      at process.processImmediate (node:internal/timers:476:21)
    ```

cenfun:
@bcoe It seems that there are some legacy files on your local, possibly generated from a previous running but not test. can you try removing the `coverage` dir and test again?

bcoe (approved):
LGTM! thank you for all the hard work getting this ready to land.

PR #537, opened 2024-06-12
Make monocart-coverage-reports an optional peer dep

jakebailey:
I noticed in https://github.com/bcoe/c8/commit/13979a76b5b44fc6758f350bae4cb3febd60d75e that you fixed it being a "required" peer dep by not declaring it at all.

Package managers like `pnpm` (or others) may use a stricter `node_modules` layout which would prevent `c8` from accessing `monocart-coverage-reports` because it's not a declared dependency.

Instead of not declaring the dep, you can still use a peer dep but just mark it optional. Users installing `c8` will not get `monocart-coverage-reports` automatically, which I believe is what you intended. But, installing it later will ensure that it's correctly linked.

As an example, see https://github.com/microsoft/TypeChat/blob/ce83b7f63f19786106f82a37ca154bbb579d960b/typescript/package.json#L43-L54, which declares deps on libraries that can plug into TypeChat, but are not hard deps.

(I also find this to more clearly show intent to people reading the package or analyzing their deps.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,8 +56,13 @@
     "ts-node": "^10.7.0",
     "typescript": "^5.0.0"
   },
-  "c8Requirements": {
+  "peerDependencies": {
     "monocart-coverage-reports": "^2"
+  },
+  "peerDependenciesMeta": {
+    "monocart-coverage-reports": {
+      "optional": true
+    }
   },
   "engines": {
     "node": ">=18"
--- a/package.json
+++ b/package.json
@@ -51,7 +51,6 @@
     "chai-jest-snapshot": "^2.0.0",
     "cross-env": "^7.0.3",
     "mocha": "^9.2.2",
-    "monocart-coverage-reports": "^2.8.3",
     "standard": "^16.0.4",
     "ts-node": "^10.7.0",
     "typescript": "^5.0.0"
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,6 +37,14 @@
       },
       "engines": {
         "node": ">=18"
+      },
+      "peerDependencies": {
+        "monocart-coverage-reports": "^2"
+      },
+      "peerDependenciesMeta": {
+        "monocart-coverage-reports": {
+          "optional": true
+        }
       }
     },
     "node_modules/@aashutoshrathi/word-wrap": {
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -51,14 +51,18 @@
     "chai-jest-snapshot": "^2.0.0",
     "cross-env": "^7.0.3",
     "mocha": "^9.2.2",
-    "monocart-coverage-reports": "^2.8.3",
     "standard": "^16.0.4",
     "ts-node": "^10.7.0",
     "typescript": "^5.0.0"
   },
-  "c8Requirements": {
+  "peerDependencies": {
```
bcoe, line 58: I believe this will install by default, and what I'm trying to avoid is forcing an 800kb for people who are not experimenting with monocart.

jakebailey: It won't, I promise! As an example, I did `npm install typechat`:

```console
$ npm install typechat

added 1 package, and audited 2 packages in 1s

found 0 vulnerabilities
$ ls node_modules
typechat
$ cat package-lock.json
{
  "name": "typechattest",
  "version": "1.0.0",
  "lockfileVersion": 3,
  "requires": true,
  "packages": {
    "": {
      "name": "typechattest",
      "version": "1.0.0",
      "license": "ISC",
      "dependencies": {
        "typechat": "^0.1.0"
      }
    },
    "node_modules/typechat": {
      "version": "0.1.0",
      "resolved": "https://registry.npmjs.org/typechat/-/typechat-0.1.0.tgz",
      "integrity": "sha512-1QeypvhyoXz/1DssDq8NbIzRHX3V20b+TWfrO7t4wd3o8OygcQO3Zo9MRNSBwK0Pyw+Fw6MmE/hOFi372gpvzA==",
      "workspaces": [
        "./",
        "./examples/*"
      ],
      "engines": {
        "node": ">=18"
      },
      "peerDependencies": {
        "typescript": "^5.3.3",
        "zod": "^3.22.4"
      },
      "peerDependenciesMeta": {
        "typescript": {
          "optional": true
        },
        "zod": {
          "optional": true
        }
      }
    }
  }
}
```

jakebailey: (This is what `peerDependenciesMeta` with `optional` is intended to handle.)

bcoe: Huh, cool I will test too.

bcoe (approved):
Thank you for the contribution.

Issue #498, opened 2023-10-31
c8 CLI doesn't work properly when executed in a sub-directory of the project


ericmorand:
* **Version**: 8.0.1
* **Platform**: Ubuntu 20.04 + Node.js 16.20.2

Consider the following project:

#### src/index.ts
```ts
export const foo = (bar: number) => {
    return bar + 1;
};
```

#### test/index.ts
```ts
import assert from "assert";
import {foo} from "../src";

assert(foo(5) === 6);
```

Depending on where `c8` is executed, the result is quite different:

#### Executed from the root of the project
```sh
$ c8 ts-node test/index.ts 

=============================== Coverage summary ===============================
Statements   : 100% ( 3/3 )
Branches     : 100% ( 1/1 )
Functions    : 100% ( 1/1 )
Lines        : 100% ( 3/3 )
================================================================================
```

#### Executed from the place where the test file is (i.e. `test`)
```sh
$ cd test/
$ c8 ts-node index.ts 

=============================== Coverage summary ===============================
Statements   : Unknown% ( 0/0 )
Branches     : Unknown% ( 0/0 )
Functions    : Unknown% ( 0/0 )
Lines        : Unknown% ( 0/0 )
================================================================================
```

For information - and maybe it can help - here is the `.nycrc.json` that I use:

```json
{
  "all": true,
  "lines": 100,
  "branches": 100,
  "functions": 100,
  "extension": [
    ".ts"
  ],
  "include": [
    "src"
  ],
  "exclude": [],
  "reporter": [
    "text-summary",
    "html"
  ]
}
```

Let me also add that `nyc` works perfectly regardless of where it is executed - i.e. replace `c8` with `nyc` in the showcased shell commands and the report is perfect in both case. We educated all our developers too execute coverage as soon as possible and on local tests, and unfortunately `c8` not supporting this pattern means that we can't migrate to it from `nyc`.

Hopefully this is solvable.

skullpsg:
Can you try with --allowExternal on the cmd. It would work.

ericmorand:
I'll try but I think it won't: c8 resolves the `src` reference from the working directory and will then expect that the files exist in `test/src`.

ericmorand:
For information, after working on PR #539 and seeing how fixing the current issue would break c8, I created a code coverage tool from scratch that is not affected by the issue : https://www.npmjs.com/package/one-double-zero. Please give it a try if you are affected by the current issue.

PR #539, opened 2024-06-23
Fix issue #498 - c8 CLI doesn't work properly when executed in a sub-directory of the project

ericmorand:
This PR fixes the issue #498...and also change the philosophy of c8.

### The fix

This PR makes c8 consider the `include` and `exclude` patterns as relative to the configuration file location, if any, or the current working directory, otherwise. By applying such a change, c8 can be executed from any subdirectory of the project and will always resolve sources to include and exclude regardless of the current working directory, if a configuration file is present.

### The philosophy shift

c8 philosophy seems to have been heavily inspired by nyc, which means that it inherits from nyc debatable (and arguably archaic) design decisions. These decisions can be summarized this way:

Options `src`, `include`, `exclude`, `extension`, `excludeNodeModules`, `allowExternal` and `all` are overlapping and can be reduced to only `include` and `exclude` without removing any feature.

This PR removes all the mentioned options except `include` and `exclude`, and updates the logic so that all the removed options are organically available through the `include` and `exclude` options.

Let's see how it goes.

#### --src

This option is overlapping with the `include` option, which _is_ the declaration of the files that we want to be covered by the tests:

```
--include=src/main/**/*.js
```

#### --extension

This option is overlapping with the `include` option, which _is_ the declaration of the files that we want to be covered by the tests:

```
--include=src/main/**/*.ts --include=src/main/**/*.tsx
```

Or:

```
--include=src/main/**/*.{ts,tsx}
```

#### --allowExternal

This option is overlapping with the `include` option, which _is_ the declaration of the files that we want to be covered by the tests:

```
--include=src/main/**/*.ts --include=../../../../external-project/*.js
```

#### --excludeNodeModules

This option is overlapping with the `exclude` option, which _is_ the declaration of the files that we don't want to be covered by the tests:

```
--include=**/*.js --exclude=node_modules/**/*
```

#### --all

This option is overlapping with the `include` and `exclude` options, would we want to not include all of them, it is possible through a combination of `include` and `exclude`:

```
--include=**/*.js --exclude=**/not-included/*.js
```

This is obviously a breaking change.

### DIFF
--- a/test/integration.js
+++ b/test/integration.js
@@ -120,23 +120,19 @@ beforeEach(function () {
       output.toString('utf8').should.matchSnapshot()
     })
 
-    it('should allow for multiple overrides of src location for --all', () => {
+    it('should allow for multiple include patterns', () => {
       // Here we nest this test into the report directory making the multidir
       // directories outside of cwd. Note, that the target srcOverride does not
-      // require fields from these directories but we want them initialized to 0
-      // via --all. As such we --allowExternal and provide multiple --src patterns
-      // to override cwd.
+      // require fields from these directories.
       const { output, status } = spawnSync(nodePath, [
         c8Path,
         '--exclude="test/*.js"',
         '--temp-directory=../tmp/src',
         '--clean=true',
-        '--allowExternal',
         '--reporter=text',
-        '--all',
-        `--src=${dirname(require.resolve('./fixtures/multidir1/file1.js'))}`,
-        `--src=${dirname(require.resolve('./fixtures/multidir2/file2.js'))}`,
-        `--src=${dirname(require.resolve('./fixtures/report/srcOverride.js'))}`,
+        '--include="./test/fixtures/multidir1/*.js"',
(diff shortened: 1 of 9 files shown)


test/integration.js
```diff
@@ -499,100 +496,6 @@ beforeEach(function () {
         output.toString('utf8').should.matchSnapshot()
       })
     })
-    describe('--all', () => {
```
bcoe, line 502: I'm concerned about removing so many tests in this PR.

bcoe:
@ericmorand I understand where you're coming from, but this breaking change would be pretty drastic, and I don't have the time to put into open source these days to support it.

Could we instead figure out how to make `--allowExternal` work for your case, and fix any bugs you bump into.

ericmorand:
@bcoe , your point is fair and I actually agree that this is too much of a drastic change, even for a breaking change: this is like a different product. I'll probably end up proposing a dedicated package - heavily inspired by your great work. I think we can close this PR.

PR #560, opened 2025-04-29
chore(deps): replace find-up with empathic

### LINKED ISSUE #559: Replace `find-up` with smaller alternatives
`find-up` has a few dependencies and there are some smaller alternatives:

- [`find-up-simple`](https://github.com/sindresorhus/find-up-simple) is also made by sindre and is intended to be a smaller version of the original
  - however, it only supports a single filename input instead of the [array that this package uses](https://github.com/bcoe/c8/blob/1ec3cc41ff1aeb59bb040ee9c143a09054ac413b/lib/parse-args.js#L21), so it would require some extra code
- [`empathic`](https://github.com/lukeed/empathic) is made by lukeed and is a teeny utility library for paths
  - `find.any()` could replace the usage in this package
  - but the package is ESM-only, ~~and would require a little dance to use until the package can rely on `require(esm)` in node 22+~~:
    ```js
    default: async () => import('empathic/find').then(({ any }) =>
      any(['.c8rc', '.c8rc.json', '.nycrc', '.nycrc.json']))
    ```
    `default` does not accept async functions so this won't work 🙁 

---

if you are interested I could also open an issue for replacing the quite large `test-exclude` as well!

beeequeue:
This PR replaces `find-up` with the smaller and faster alternative `empathic`, now that it has been ported to cjs!

This will clean up the following branch of transitive dependencies into just one:

<details>
<summary>Details</summary>

image

</details>

##### Checklist

closes #559

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node: [18, 20]
+        node: [18, 20, 22]
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
@@ -27,7 +27,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
-          node-version: v20.18.1 # TODO: switch to lts/* once https://github.com/nodejs/node/issues/51766 lands.
+          node-version: lts/*
           cache: npm
       - run: npm ci
       - run: npm test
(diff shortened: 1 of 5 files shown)


bcoe:
Hey @beeequeue I appreciate the PR, but find-up is only `1.6kb` over the wire:

Screenshot 2025-05-02 at 11 57 45 AM

I don't see value in swapping out a dependency that has been battle tested for years with thousands of dependencies, vs., trying something new (even though I'm sure it's great).

43081j:
@bcoe can we revisit this?

it isn't just about the size, although it is smaller. it is actually faster and, more importantly replaces multiple packages, not only `find-up`.

the "multiple packages" part doesn't affect you directly here, but it cleans a lot up in consuming packages and seems a shame that they then pull the slower package in anyway because they depend on `c8`.

replacing it here could be a very helpful stepping stone to replacing it in _many_ places above this package.

let me know what you think

PR #578, opened 2026-02-22
fix(deps): update yargs to 18

bcoe:
Now that we've bumped the major due to a recent CVE, might as well update yargs to version with less build steps.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,7 @@
         "istanbul-reports": "^3.1.6",
         "test-exclude": "^8.0.0",
         "v8-to-istanbul": "^9.0.0",
-        "yargs": "^17.7.2",
+        "yargs": "^18.0.0",
         "yargs-parser": "^21.1.1"
       },
       "bin": {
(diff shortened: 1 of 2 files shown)


coderaiser:
Would be great also to update yargs-parser 

https://github.com/yargs/yargs-parser/releases/tag/yargs-parser-v22.0.0

bricss:
Perhaps `yargs` can be replaced with the native [util.parseArgs([config])](https://nodejs.org/api/util.html#utilparseargsconfig), which has been stable in Node since `v20` 🤓

coderaiser:
Looks like this can be uncommented:

https://github.com/bcoe/c8/blob/main/lib/parse-args.js#L176-L187
https://github.com/bcoe/c8/blob/main/lib/commands/report.js#L13

JoernBerkefeld:
to support that, yargs 17 has an issue with node 25.7 (also see #581) which is fixed in yargs 18. do please update that dependecy


file:///home/runner/work/SFMC-SDK/SFMC-SDK/node_modules/yargs/yargs:3
const {applyExtends, cjsPlatformShim, Parser, Yargs, processArgv} = require('./build/index.cjs')
                                                                    ^

ReferenceError: require is not defined in ES module scope, you can use import instead
    at file:///home/runner/work/SFMC-SDK/SFMC-SDK/node_modules/yargs/yargs:3:69
    at ModuleJobSync.runSync (node:internal/modules/esm/module_job:534:37)
    at ModuleLoader.importSyncForRequire (node:internal/modules/esm/loader:366:47)
    at loadESMFromCJS (node:internal/modules/cjs/loader:1628:24)
    at Module._compile (node:internal/modules/cjs/loader:1793:5)
    at Object..js (node:internal/modules/cjs/loader:1951:10)
    at Module.load (node:internal/modules/cjs/loader:1532:32)
    at Module._load (node:internal/modules/cjs/loader:1334:12)
    at wrapModuleLoad (node:internal/modules/cjs/loader:255:19)
    at Module.require (node:internal/modules/cjs/loader:1555:12)

Node.js v25.7.0
Error: Process completed with exit code 1.

bricss:
The import issue has been resolved -> https://github.com/bcoe/c8/pull/581#issuecomment-4047192632 🩹

JoernBerkefeld:
While that's good news, getting c8  upgraded to use yargs 18 is still a good idea

shivarm:
@bcoe check #592 and please can you merge it ?

bricss:
Working solution -> https://github.com/yargs/yargs/issues/2509#issuecomment-4412569569

package-lock.json
```diff
@@ -4787,20 +4978,20 @@
...
-        "require-directory": "^2.1.1",
-        "string-width": "^4.2.3",
+        "string-width": "^7.2.0",
         "y18n": "^5.0.5",
-        "yargs-parser": "^21.1.1"
+        "yargs-parser": "^22.0.0"
       },
       "engines": {
-        "node": ">=12"
+        "node": "^20.19.0 || ^22.12.0 || >=23"
```
ericcornelissen, line 4994: Technically, this is more restrictive than `c8`'s supported Node.js version range

https://github.com/bcoe/c8/blob/ee2f1cfc5584d41bb2d51b788d0953dab0c798f8/package.json#L67

and hence would require a major version bump.

This is the strictest version range I could find (shared with [`yargs-parser`](https://github.com/bcoe/c8/pull/578/changes#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519R5076))

bricss:
It would be very nice ✨ to have this merged ⛜ and a new 🆕 version 🚀 released 📦, as `c8` is failing 💥 right now ⏰ all over the ecosystem 🌍.

package.json
```diff
@@ -42,7 +42,7 @@
     "istanbul-reports": "^3.1.6",
     "test-exclude": "^8.0.0",
     "v8-to-istanbul": "^9.0.0",
-    "yargs": "^17.7.2",
+    "yargs": "^18.0.0",
```
andersk, line 45: What if we declare compatibility with both versions, to avoid the need for a major version bump pointed out in https://github.com/bcoe/c8/pull/578#discussion_r3213229107?

```suggestion
    "yargs": ">=17.7.2 && <19"
```

remcohaszing: This won’t work. This is a version range. Within this range, package managers will try to resolve the newest version. At the time of writing, the latest version that satisfies either of these ranges, is 18.0.0.

andersk: What is happening here, then?

```console
$ echo '{"dependencies": {"yargs": ">=17.7.2 && <19"}}' > package.json
$ mise exec node@20.18.0 -- npm i
$ jq .version node_modules/yargs/package.json 
"17.7.2"
$ rm -rf node_modules package-lock.json
$ mise exec node@20.19.0 -- npm i
$ jq .version node_modules/yargs/package.json 
"18.0.0"
$ rm -rf node_modules package-lock.json
$ mise exec node@22.11.0 -- npm i
$ jq .version node_modules/yargs/package.json 
"17.7.2"
$ rm -rf node_modules package-lock.json
$ mise exec node@22.12.0 -- npm i
$ jq .version node_modules/yargs/package.json 
"18.0.0"
```

bricss: It's `npm` is respecting the package's 📦 `engines.node` constraint and choosing the newest compatible version.

Starting with `yargs` 18.0.0, the package declares:

```json
"engines": {
  "node": "^20.19.0 || ^22.12.0 || >=23"
}
```

| Node version | Compatible with `yargs` 18? | Installed |
| ------------ | ------------------------- | --------- |
| 20.18.0      | ❌                         | 17.7.2    |
| 20.19.0      | ✅                         | 18.0.0    |
| 22.11.0      | ❌                         | 17.7.2    |
| 22.12.0      | ✅                         | 18.0.0    |

Marvelous 🌟

bricss:
How-to resolve module issue w/o `npm` overrides:
```shell
npm install github:bcoe/yargs#yargs-18 --save-dev
```

package.json
remcohaszing: Apparently I was wrong about this.

I find this a bit odd though. npm will write it to the lockfile. On subsequent installs, npm has to abide by the lockfile, which most of the time has been generated using a different version of Node.js.

ericcornelissen:
The need for this has been reduced with the release of [v17.7.3](https://github.com/yargs/yargs/releases/tag/v17.7.3), see https://github.com/yargs/yargs/issues/2509#issuecomment-4747961360

PR #589, opened 2026-04-01
fix: include uncovered files in report when using `--all` with `--include`

### LINKED ISSUE #588: Coverage report is missing uncovered files when using `--all` with `--include`
When running the test suite of **MySQL2** with **c8** using `--all` combined with `--include`, some files that exist and match the include patterns become **invisible** in the coverage report when totally uncovered.

For context, here is the coverage configuration used:

```json
{
  "all": true,
  "include": ["index.js", "promise.js", "lib/**/*.js"],
  "exclude": ["mysqldata/**", "node_modules/**", "test/**"],
  "reporter": ["text", "lcov", "cobertura"],
  "statements": 80,
  "branches": 80,
  "functions": 77,
  "lines": 80,
  "checkCoverage": true,
  "clean": true
}
```

However, `lib/results_stream.js` and `lib/packets/binlog_query_statusvars.js` don't appear in the report.

**Expected behavior:**

When bypassing **c8**'s filtering (e.g., collecting **V8** coverage data directly and calling `Report` programmatically) from my [custom plugin that uses **C8** behind the scenes](https://github.com/pokujs/c8), both files correctly appear with 0% coverage, confirming they exist and match the patterns.

> [!NOTE]
> - [x] I'm preparing a fix for this 🙋🏻‍♂️

wellwelwel:
##### Checklist

---

Fixes #588.

In `lib/report.js`, the `_normalizeProcessCov` method adds files to `fileIndex` **before** checking `shouldInstrument()`:

```js
if (/^file:\/\//.test(v8ScriptCov.url)) {
  // ...
    v8ScriptCov.url = fileURLToPath(v8ScriptCov.url)
    fileIndex.add(v8ScriptCov.url) //    added unconditionally
  // ...
if ((!this.omitRelative || isAbsolute(v8ScriptCov.url))) {
  if (this.excludeAfterRemap || this._shouldInstrument(v8ScriptCov.url)) {
    result.push(v8ScriptCov) //          only added to result if passes filter
  }
}
```

Then:

- **V8** loads, for example, `lib/results_stream.js` during execution (via `require()` chain)
- `fileIndex.add(v8ScriptCov.url)` is called unconditionally
- `shouldInstrument()` returns `false`, so the file is **not** added to `result`
- `_includeUncoveredFiles` finds the file via glob, but `fileIndex.has()` is `true`, so it's **skipped** ❌
- The file is absent from both **V8** results and uncovered files

---

The fix moves `fileIndex.add()` to inside the `shouldInstrument` block, so only files that actually pass the filter are tracked.

---

### About the test approach

The test uses `createReport` and `_normalizeProcessCov` directly instead of the `spawnSync` + `matchSnapshot` pattern used in most integration tests.

Since `test-exclude`'s `globSync` internally calls `shouldInstrument` on every file it returns, the paths that fail `shouldInstrument` in `_normalizeProcessCov` won't collide with the paths from `globSync` in `_includeUncoveredFiles`. Because of that, a regular integration test would pass with or without the fix, so I couldn't find a way to catch a regression through the **CLI** alone, for example: https://github.com/wellwelwel/c8/pull/1#issue-4182866442.

By calling `_normalizeProcessCov` directly with fake **V8** coverage entries, we can verify that `fileIndex` is only populated with files that pass `shouldInstrument`.

> [!NOTE]
> I can also move this test to a separate file for unit tests if that makes more sense 🙋🏻‍♂️


wellwelwel:
Just as a proof of concept, here is the current test failure without the fix applied:

```sh
2 failing

1) c8
     --all
       should only track files that pass shouldInstrument in fileIndex:

    AssertionError: expected true to equal false
    + expected - actual

    -true
    +false
    
    at Context.<anonymous> (test/integration.js:543:44)
    at process.processImmediate (node:internal/timers:504:21)

2) c8 mergeAsync
     --all
       should only track files that pass shouldInstrument in fileIndex:

    AssertionError: expected true to equal false
    + expected - actual

    -true
    +false
    
    at Context.<anonymous> (test/integration.js:543:44)
    at process.processImmediate (node:internal/timers:504:21)
```

bcoe:
@wellwelwel apologies for ignoring this for months. Is this still something you'd like to land?

wellwelwel:
> @wellwelwel apologies for ignoring this for months. Is this still something you'd like to land?

Hey! Yes, I do. If you need any changes, I'm here 🙋🏻‍♂️

